PR #7, opened 2022-10-13
Implement Netlify runtime integration

RichardAkman:


ts/packages/cli/src/helpers/synchronizers/logic/Netlify.ts
```diff
@@ -0,0 +1,36 @@
...
+const label = CONFIG_SYNCHRONIZER_LABEL.Netlify;
+
+export type NetlifyConfiguration = {
+  // * Netlify docs: https://docs.netlify.com/configure-builds/environment-variables/
+  token: string;
+  siteId: string;
+};
+
+// TODO - Decide between syncNetlifyConfigs (previous name) and syncNetlifyEnvVars
+export const syncNetlifyEnvVars = async ({
```
RichardAkman, line 14: @rannn505 We previously named this `syncNetlifyConfigs`, Netlify names their 'configs' as 'environment variable' so I thought `syncNetlifyEnvVars` would be more appropriate here. Opinion?

PR #8, opened 2022-10-13
Implement GCP Cloud Functions runtime integration

RichardAkman:


ts/packages/cli/src/helpers/synchronizers/logic/GcpCloudFunctions.ts
```diff
@@ -0,0 +1,32 @@
...
+const label = CONFIG_SYNCHRONIZER_LABEL.GcpCloudFunctions;
+
+export type GcpCloudFunctionsConfiguration = {
+  // * GCP Cloud Functions docs: https://cloud.google.com/functions/docs/configuring/env-var
+  keyFile: string;
+  functionName: string;
+};
+
+// TODO - Decide between syncGcpCloudFunctionsConfigs (previous name) and syncGcpCloudFunctionsEnvVars
+export const syncGcpCloudFunctionsEnvVars = async ({
```
RichardAkman, line 14: @rannn505 We previously named this `syncGcpCloudFunctionsConfigs`, GCP cloud functions names their 'configs' as 'runtime environment variable' so I thought `syncGcpCloudFunctionsEnvVars` would be more appropriate here. Opinion?

RichardAkman: We decided on `syncGcpCloudFunctionsEnvironmentVariables`.

rannn505 (commented):
LGTM

PR #9, opened 2022-10-13
Implement Vercel runtime integration

RichardAkman:


ts/packages/cli/src/helpers/synchronizers/logic/Vercel.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export type VercelConfiguration = {
+  // * Vercel docs: https://devcenter.heroku.com/articles/platform-api-reference#config-vars
+  token: string;
+  appUid: string;
+};
+type VercelEnvVar = { key: string; id: string; [otherKeys: string]: unknown };
+
+// TODO - Decide between syncVercelConfigs (previous name) and syncVercelEnvVars
+export const syncVercelEnvVars = async ({
```
RichardAkman, line 20: @rannn505 We previously named this `syncVercelConfigs`, Vercel names their 'configs' as 'environment variable' so I thought `syncVercelEnvVars` would be more appropriate here. Opinion?

RichardAkman: We agreed on `syncVercelEnvironmentVariables`.

PR #10, opened 2022-10-13
Implement Azure Functions runtime integration

RichardAkman:


ts/packages/cli/src/helpers/synchronizers/logic/AzureFunctions.ts
```diff
@@ -0,0 +1,45 @@
...
+  tenantId: string;
+  clientId: string;
+  clientSecret: string;
+  subscriptionId: string;
+  resourceGroupName: string;
+  functionApp: string;
+};
+
+// TODO - Decide between syncAzureFunctionsConfigs (previous name) and syncAzureFunctionsAppSettings
+export const syncAzureFunctionsAppSettings = async ({
```
RichardAkman, line 19: @rannn505 We previously named this `syncAzureFunctionsConfigs`, Azure Functions names their 'configs' as 'application settings' so I thought `syncAzureFunctionsAppSettings` would be more appropriate here. Opinion?

RichardAkman: We agreed on `syncAzureFunctionsAppSettings`.

PR #25, opened 2022-10-31
Refactor HashiCorp vault to use kv store

RichardAkman:


ts/packages/node/src/stores/HashiCorpVault.ts
```diff
@@ -21,7 +21,7 @@ export class HashiCorpVaultStore extends Store {
   }
 
   // * K/V2 Read Secret Version - https://www.vaultproject.io/api-docs/secret/kv/kv-v2#read-secret-version
-  private getSecretPath({ set, schema }: StoreQuery[number]) {
+  calcKey({ set, schema }: StoreQuery[number]): string {
     // * set first element used as vault engine
     if (!set) {
       throw new Error(`invalid secret path is at ${this.constructor.name}`);
```
RichardAkman, line 27: @rannn505 Getting the following error here:
`Cannot read properties of undefined (reading 'constructor')`
Would like your input here.

PR #26, opened 2022-11-09
Store logic refactor

rannn505:
- Set jtd
- Store logic refactors
- KeyValueStore
- URI utils


RichardAkman:
Fixes some issues. LGTM after playing around with a local refactored AWS store.

PR #32, opened 2022-11-13
Implement Azure Key Vault store

RichardAkman:


ts/packages/node/src/stores/AzureKeyVault.ts
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  // * https://learn.microsoft.com/en-us/javascript/api/@azure/keyvault-secrets/secretclient?view=azure-node-latest#@azure-keyvault-secrets-secretclient-setsecret
+  async upsert(key: string, value: string): Promise<void> {
+    await this.client.setSecret(key, JSON.stringify(value));
+  }
+
+  // * https://learn.microsoft.com/en-us/javascript/api/@azure/keyvault-secrets/secretclient?view=azure-node-latest#@azure-keyvault-secrets-secretclient-setsecret
+  async delete(key: string): Promise<void> {
+    await this.client.beginDeleteSecret(key);
```
RichardAkman, line 34: @rannn505 

`beginDeleteSecret` is a soft delete method that will prevent future upserts from working until the scheduled deletion occurs.

As it stands, the current implementation is an unpleasant UX. I propose any of the following options:
1) Immediately use the `purgeDeletedSecret` method to remove the secret and allow upserts. This logic is in line with how the rest of the stores work.
2) Give the user an option to decide if he wants a 'soft' or 'hard' delete. This will give the user fine-grained control but will cost us a more complicated implementation on both the delete and upsert logic.
3) Keep everything as is and document the specific behavior in the docs. The user needs to manually delete the secret from the interface before he can upsert

RichardAkman: @rannn505 

I tried to implement and test this logic and discovered that in order to use the `purgeDeletedSecret` method, we need to use a poller that is returned from `beginDeleteSecret` method and `await` on it. The problem is that it takes at least 30 to 60 seconds for the operation to complete and it hangs the command for an unreasonable amount of time. What do you think about this situation?

ts/packages/node/src/stores/AzureKeyVault.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  // * https://learn.microsoft.com/en-us/javascript/api/@azure/keyvault-secrets/secretclient?view=azure-node-latest#@azure-keyvault-secrets-secretclient-setsecret
+  async upsert(key: string, value: string): Promise<void> {
+    await this.client.setSecret(key, value);
+  }
+
+  // * https://learn.microsoft.com/en-us/javascript/api/@azure/keyvault-secrets/secretclient?view=azure-node-latest#@azure-keyvault-secrets-secretclient-begindeletesecret
+  async delete(key: string): Promise<void> {
+    await this.client.beginDeleteSecret(key);
```
rannn505, line 38: add comment with // ! explain the limitation

rannn505 (approved):

ts/packages/node/src/stores/AzureKeyVault.ts
RichardAkman: done

ts/packages/node/src/stores/AzureKeyVault.ts
RichardAkman: @rannn505 & I decided to leave the logic as is as a limitation.

PR #36, opened 2022-11-15
Add store initialization logic to commands

RichardAkman:


ts/packages/ts/src/commands/UpsertCommand.ts
```diff
@@ -72,6 +72,8 @@ export class UpsertCommand extends Command<void> {
       })
       .value();
 
+    await store.init();
```
rannn505, line 75: this should be called at the beginning of the command

ts/packages/ts/src/commands/EvalCommand.ts
```diff
@@ -38,6 +38,10 @@ export class EvalCommand extends Command<EvalCommandReturn> {
         })
         .value();
     });
+
+    // * initialize all provided stores
+    await Promise.all(stores.map((storeInstance) => storeInstance.init()));
```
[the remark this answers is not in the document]
RichardAkman: Done

ts/packages/ts/src/commands/UpsertCommand.ts
RichardAkman: Done

rannn505 (approved):
LGTM

PR #43, opened 2022-11-16
Change CLI URL usage to URI

RichardAkman:


ts/packages/cli/src/helpers/stores/index.ts
```diff
@@ -1,26 +1,24 @@
...
+  ...JsonFileStoreSTI,
+  ...HashiCorpVaultStoreSTI,
+  ...AwsSecretsManagerStoreSTI,
 };
 
-export const constructStoreFromUrl = (url: string) => {
-  const storeUrl = new URL(url);
-  const storeProtocol = storeUrl.protocol.slice(0, -1);
-  const storeInitFunction = PROTOCOL_TO_STORE_INIT_FN_DICT[storeProtocol];
+export const constructStoreFromUri = (uri: string) => {
```
RichardAkman, line 17: @rannn505 

The change we made causes us to duplicate the following lines for all stores:
``` node.js
    const parsedUri = URI.parse(uri);
    const queryDict = _.fromPairs(parsedUri.query?.split('&').map((query) => query.split('=')));
    const splittedUserinfo = parsedUri.userinfo?.split(':');
```

I suggest that we adjust the `SchemeToInit` logic to supply these details directly to each store to prevent duplication. If we do so then we need to adjust the @configu/ts to export the `URIComponents` type. Should we do it now or open an issue for this for later?

rannn505 (approved):

ts/packages/cli/src/helpers/stores/index.ts
RichardAkman: @rannn505 & I decided to tackle this now.

PR #44, opened 2022-11-16
Implement TypeORM stores

RichardAkman:


RichardAkman:
We had to refactor and adjust the jest-related logic in order to get the tests to work properly with the decorators that the TypeORM package requires.

ts/packages/node/src/stores/Postgres.ts
```diff
@@ -0,0 +1,10 @@
+import { PostgresConnectionOptions } from 'typeorm/driver/postgres/PostgresConnectionOptions';
```
rannn505, line 1: Rename to PostgreSQL

ts/packages/node/tsconfig.json
```diff
@@ -3,11 +3,9 @@
   "compilerOptions": {
     "rootDir": "src",
     "outDir": "build",
-    "composite": true,
+    "composite": true
```
rannn505, line 6: diff is irelevant here

rannn505 (approved):

ts/packages/node/src/stores/Postgres.ts
RichardAkman: done

ts/packages/node/tsconfig.json
RichardAkman: managed to remove the diff

PR #51, opened 2022-11-18
Implement CLI support for ORM stores

RichardAkman:


ts/packages/cli/src/helpers/stores/TypeORM.ts
```diff
@@ -0,0 +1,98 @@
...
+  if (
+    (parsedUri.scheme === AuroraMysqlStore.scheme || parsedUri.scheme === AuroraPostgreSQLStore.scheme) &&
+    (!region || !secretArn || !resourceArn)
+  ) {
+    throw new Error(`invalid store uri ${uri}`);
+  }
+
+  return {
+    uri,
+    store: {
```
RichardAkman, line 34: @rannn505, I feel like there should probably be a way to use typescript to construct a new instance based on a given class generically.

ts/packages/cli/src/helpers/stores/TypeORM.ts
```diff
@@ -0,0 +1,98 @@
...
+  MSSQLStore,
+  MariaStore,
+  MySQLStore,
+  PostgreSQLStore,
+  SQLiteStore,
+} from '@configu/node';
+import { InitFunctionParameters, SchemeToInit } from './types';
+
+const STI = async ({ uri, parsedUri, userinfo, queryDict }: InitFunctionParameters) => {
+  const [username, password] = userinfo;
```
RichardAkman, line 14: @rannn505, the username, and password could potentially include characters that break the URI pattern. For example, the username may be an email address that breaks the URI because of the `@` char. How do you propose we handle this?

ts/packages/cli/src/helpers/stores/TypeORM.ts
```diff
@@ -0,0 +1,99 @@
...
+  CockroachStore,
+  MSSQLStore,
+  MariaStore,
+  MySQLStore,
+  PostgreSQLStore,
+  SQLiteStore,
+} from '@configu/node';
+import { InitFunctionParameters, SchemeToInit } from './types';
+
+const STI = async ({ uri, parsedUri, userinfo, queryDict }: InitFunctionParameters) => {
```
rannn505, line 13: STI what?

ts/packages/cli/src/helpers/stores/TypeORM.ts
RichardAkman: An issue will be opened for this.

ts/packages/cli/src/helpers/stores/TypeORM.ts
RichardAkman: renamed

rannn505 (approved):

PR #73, opened 2022-12-25
[WIP] feat: python SDK

orlevii:


ShriekinNinja (approved):
we'll use only what we'll need, thanks @orlevii

PR #74, opened 2022-12-27
CLI get-started flow

rannn505:
- init: change --name to --uid with alias
- init: add GET_STARTED populated schema data and flag
- store upsert: add --type flag, label is not required, only available with --interactive and skips type selection


RichardAkman (approved):
LGTM

PR #85, opened 2023-01-22
Fix export failure when using source flag with special chars

RichardAkman:


ts/packages/lib/src/formatters/configFormatters.ts
```diff
@@ -5,21 +5,12 @@ import validator from 'validator';
...
 
 const jsonToDotenv: FormatterFunction = ({ json }) => {
   return Object.entries(json)
-    .map(([key, value]) => {
-      if (hasWhitespace(value)) {
-        return `${key}="${value}"`; // * in case value has a whitespace, wrap with quotes around it
-      }
-      return `${key}=${value}`;
-    })
+    .map(([key, value]) => `${key}="${value}"`)
```
RichardAkman, line 13: @rannn505 instead of always wrapping the value with quotes we could decide via a given parameter called `wrapValueWithQuotes` or something like that. We can have the CLI set the param to true when using the source flag. What do you think?

RichardAkman: go for this option, call the optional param `wrap`. add comment to explain this param

ts/packages/lib/src/formatters/configFormatters.ts
```diff
@@ -9,14 +9,18 @@ const hasWhitespace = (str: string) => {
   return /\s/.test(str);
 };
 
-type FormatterParameters = { json: EvaluatedConfigs; label: string };
+type FormatterParameters = {
+  json: EvaluatedConfigs;
+  label: string;
+  wrap?: boolean; // * Wraps all values with quotes
```
RichardAkman, line 15: Maybe add comment to implement this wrap for rest of formatters?

rannn505:
@RichardAkman any reason we shouldn't merge this?

rannn505 (approved):

PR #128, opened 2023-06-20
Code scanner to find config declaration location based on .cfgu files

RonConfigu:
very immature version. just a POC.
- probably best to add optional flag for specific cfgu
- if no cfgu is found, maybe parse one from .env?
- typescript AST maybe needed for more complex tasks
- convert use of for..for.. to reduce / lodash


ts/packages/cli/src/commands/scan.ts
```diff
@@ -0,0 +1,92 @@
...
+import { ConfigSchema } from '@configu/node';
+import { createInterface } from 'readline';
+import { createReadStream } from 'fs';
+import { once } from 'events';
+import { BaseCommand } from '../base';
+
+export default class Scan extends BaseCommand<typeof Scan> {
+  private readonly _ignoreFiles: string[] = ['build', 'dist', 'node_modules']; // TODO: read .gitignore add to here
+
+  static description = `Scan source code for cfgu usage`;
```
RonConfigu, line 13: test from ide

RonConfigu, line 13:
```suggestion
sdfsdf
```

RonConfigu:
image

ts/packages/cli/package.json
```diff
@@ -112,7 +112,8 @@
     "lodash": "^4.17.21",
     "log-symbols": "4.1.0",
     "open": "8.4.2",
-    "openid-client": "^5.4.2"
+    "openid-client": "^5.4.2",
+    "fast-glob": "^3.2.12"
```
rannn505, line 116: please keep deps as the ABC order.

ts/packages/cli/src/commands/find.ts
```diff
@@ -0,0 +1,175 @@
...
+      command: `<%= config.bin %> <%= command.id %> <PATH_TO_SCAN> --cfgu <PATH_TO_CFGU>`,
+    },
+    {
+      description: `Find all unused config parameters in specific directory`,
+      command: `<%= config.bin %> <%= command.id %> <PATH_TO_SCAN> --unused`,
+    },
+  ];
+
+  static args = {
+    path: Args.directory({
```
rannn505, line 33: Currently we dont use args in the entire CLI and I wish to keep that pattern as args are not named.
Please migrate it to a `dir` flag as in the init command
image

ts/packages/cli/src/commands/find.ts
```diff
@@ -0,0 +1,175 @@
...
+    cfgu: Flags.file({
+      description: 'Path to a specific .cfgu file (default: all CFGU files in the source code)',
+      required: false,
+    }),
+    unused: Flags.boolean({
+      description: 'Only show unused parameters',
+      required: false,
+      default: false,
+    }),
+    'show-templates': Flags.boolean({
```
rannn505, line 50: I think it is pretty wired to have a `show-templates` that actually means ignore something.
We can do something with the allowNo option for boolean flags as in the empty flag in the export command.
image
And then you have --templates and --no-templates.

ts/packages/cli/src/commands/find.ts
```diff
@@ -0,0 +1,175 @@
...
+      default: '.',
+    }),
+  };
+
+  static flags = {
+    cfgu: Flags.file({
+      description: 'Path to a specific .cfgu file (default: all CFGU files in the source code)',
+      required: false,
+    }),
+    unused: Flags.boolean({
```
rannn505, line 45: I think the description lacks a bit as I really dont understand form it what the purpose of this flag @pelegpor.

ts/packages/cli/src/commands/find.ts
```diff
@@ -0,0 +1,175 @@
...
+      description: 'Only show unused parameters',
+      required: false,
+      default: false,
+    }),
+    'show-templates': Flags.boolean({
+      description: 'Ignore parameters that are parts of templates and treat them as used parameters',
+      required: false,
+      default: false,
+    }),
+    ignore: Flags.string({
```
rannn505, line 55: maybe instead of Comma separated just use the multiple option and the the user can do --ignore "" --ignore "" etc.. also I wonder if we need another flag for .ignore files like .gitignore .npmignore etc..

ts/packages/cli/src/commands/find.ts
```diff
@@ -0,0 +1,175 @@
...
+      description: 'Comma separated list of glob patterns to ignore. (ex. --ignore "*.xml,**/*.md")',
+      required: false,
+      default: '',
+    }),
+  };
+
+  private async updateIgnoredFiles(projectPath: string) {
+    this.ignoredFiles = new Set([...this.ignoredFiles, ...this.flags.ignore.split(',')]);
```
rannn505, line 63: This method can return this Set instead of defining it on the class.

ts/packages/cli/src/commands/find.ts
```diff
@@ -0,0 +1,175 @@
...
+    const gitIgnores = FastGlob.sync([path.join(projectPath, '.gitignore')], {
+      dot: true,
+      ignore: [...this.ignoredFiles].map((dirName) => path.join(projectPath, `/**/${dirName}`)),
+    });
+    await Promise.all(
+      gitIgnores.map((gitIgnore) =>
```
rannn505, line 69: for better readability of the code can you do something like cons gitIgnorePromises = [...] and the await Promise.all(gitIgnorePromises)?

ts/packages/cli/src/commands/find.ts
```diff
@@ -0,0 +1,175 @@
...
+    if (unusedKeys.length > 0) this.log(`Unused configs found: ${unusedKeys.join(', ')}`);
+    else if (unusedOnly) this.log('No unused configs found');
+  }
+
+  public async run(): Promise<void> {
+    try {
+      await this.find();
+    } catch (error) {
+      this.log(`ERROR: ${error.message}`);
+    }
```
rannn505, line 173: Couple of suggestion here:
- find is too general name
- The log ERROR .. in the end is not really necessary 
  - we actually want to exit with code 1 when something went wrong
  - throwing an error actually logs an "Error" prefix as in the other commands
- Im suggesting moving all the code from find method to the run method without wrapping it in try catch

rannn505 (changes requested):
@RonConfigu please review the comments I left and let me know if everything is understood.
Regarding the flags / args thing I think we should go with
- `--dir` for the path of the find to start look from
- `--include (multiple)` for narrowing the search to specific cfgu files or dir
- `--exclude (multiple)` instead of the ignore
- `--unused` with a better description
- `--templates/--no-templates`
Also can you please attach in the next comments the output for all of the scenarios that affect it (like with no flags, with the --templates and --unused each, both of them, etc.)

Thanks 🥇

ts/packages/cli/package.json
RonConfigu: sure

ts/packages/cli/src/commands/find.ts
RonConfigu: sure

ts/packages/cli/src/commands/find.ts
RonConfigu: sure

ts/packages/cli/src/commands/find.ts
RonConfigu: sure

ts/packages/cli/src/commands/find.ts
RonConfigu: sure

RonConfigu:
image
image
image
image
image

ts/packages/cli/src/commands/find.ts
```diff
@@ -0,0 +1,165 @@
...
+    templates: Flags.boolean({
+      description: '[default: false] Show parameters that are parts of templates and treat them as used parameters',
+      required: false,
+      default: false,
+      allowNo: true,
+    }),
+  };
+
+  private async updateIgnoredFiles(projectPath: string) {
+    const ignoredFiles = new Set(['build', 'dist', 'node_modules', ...this.flags.exclude]);
```
rannn505, line 62: Suggestion for the future, we might consider using https://www.npmjs.com/package/globby that has a built in support for gitignore files

ts/packages/cli/src/commands/find.ts
```diff
@@ -0,0 +1,165 @@
...
+      count: 0,
+    }));
+
+    const files = FastGlob.sync([path.join(findInDirectory, '/**/*.*')], {
+      dot: true,
+      ignore: [...[...ignoredFiles].map((dirName) => path.join(findInDirectory, `/**/${dirName}`)), '/**/*.json'],
+    });
+
+    await Promise.all(
+      files.map((file) =>
```
rannn505, line 144: please keep the same pattern with all of the `Promise.all` as stated in the previous comments

rannn505 (approved):
@RonConfigu please fix the last comment and release

rannn505 (approved):

PR #130, opened 2023-06-22
Update the python SDK with the changes in ts/node SDKs

RonConfigu:


rannn505 (approved):
@RonConfigu looks good, i just wonder why there are diffs on the upsert command other then LGTM

PR #134, opened 2023-06-28
Fix key:value set() and remove wildcards from py

RonConfigu:


ts/packages/ts/src/stores/KeyValue.ts
```diff
@@ -14,10 +14,7 @@ export abstract class KeyValueConfigStore extends ConfigStore {
   protected abstract delete(key: string): Promise<void>;
 
   private calcKey({ set, key }: ConfigStoreQuery): string {
-    if (!set) {
-      return key;
-    }
-    return set;
+    return set || key;
```
rannn505, line 17: I know it doesnt matter but i like the strict if syntax more

ts/packages/ts/src/stores/KeyValue.ts
```diff
@@ -80,16 +77,17 @@ export abstract class KeyValueConfigStore extends ConfigStore {
   }
 
   async set(configs: Config[]): Promise<void> {
-    const kvDict: Record<string, Record<string, string>> = {};
+    const kvDict: Record<string, Record<string, string> | string> = {};
```
rannn505, line 80:
```suggestion
    const kvDict: Record<string, string | Record<string, string>> = {};
```

rannn505 (approved):

ts/packages/ts/src/stores/KeyValue.ts
RonConfigu: sure. reverted

PR #158, opened 2023-07-09
KeyValueConfigStore stringify plain strings

RonConfigu:


ts/packages/ts/src/stores/KeyValue.ts
```diff
@@ -99,7 +99,7 @@ export abstract class KeyValueConfigStore extends ConfigStore {
         return;
       }
 
-      await this.upsert(key, JSON.stringify(value));
+      await this.upsert(key, _.isPlainObject(value) ? JSON.stringify(value) : String(value));
```
RichardAkman, line 102: The following is better in my opinion:
```suggestion
      await this.upsert(key, typeof value === 'string' ? value : JSON.stringify(value));
```

This will ensure any non-string value will always get JSON.stringified instead of having `String()` convert something that `JSON.parse` might not be able to handle in some edge cases.

py/configu/stores/key_value_store.py
```diff
@@ -103,4 +103,7 @@ def set(self, configs: List[Config]):
             if value == "" or value is None:
                 self.delete(key)
             else:
-                self.upsert(key, json.dumps(value))
+                self.upsert(
+                    key,
+                    json.dumps(value) if isinstance(value, dict) else value,
```
RichardAkman, line 108: View my comment below, what happens if `value` is somehow some type other than a string and a dict? will it get properly parsed by `get`? we should make this logic airtight against unexpected edge cases.

RichardAkman (changes requested):
Good work, check out my comments.

ts/packages/ts/src/stores/KeyValue.ts
RonConfigu: I agree. Still, Ran specifically asked me to use _.isPlainObject() so please decide between you two and i'll implement here and in line 90

py/configu/stores/key_value_store.py
RonConfigu: our guard is the schema via EvalCommand and UpsertCommand. this is why everything gets stringify.
which is missing here. thanks for that. added.

RichardAkman (approved):
After reviewing the comments together in person, LGTM ⭐

PR #163, opened 2023-07-11
Python SDK refactoring

RonConfigu:


RonConfigu:
- changed max-line-length to 88 for better readability
- added DOCS.md for first page in auto-docs.
- added some docs for example
- changed api signature for `Commands` and `ConfiguConfigStore`

@jamesL92 If you have the time and will, I would love review the changes and comment 🙏

py/DOCS.md
```diff
@@ -0,0 +1,73 @@
+# @configu/py
```
rannn505, line 1: what is this file doing?

py/README.md
```diff
@@ -53,7 +51,8 @@ See [oss.configu.com/py](https://oss.configu.com/py/configu.html)
 
 ### Requirements
 
-1. Follow the [Development](https://github.com/configu/configu/blob/main/CONTRIBUTING.md#development) section from the `CONTRIBUTING.md`.
+1. Follow the [Development](https://github.com/configu/configu/blob/main/CONTRIBUTING.md#development) section from
+   the `CONTRIBUTING.md`.
```
rannn505, line 55: why newline here?

py/README.md
```diff
@@ -83,4 +82,5 @@ poetry install
 
 ### Contribute
 
-Follow the [Sending a Pull Request](https://github.com/configu/configu/blob/main/CONTRIBUTING.md#sending-a-pull-request) section from the `CONTRIBUTING.md`.
+Follow the [Sending a Pull Request](https://github.com/configu/configu/blob/main/CONTRIBUTING.md#sending-a-pull-request)
+section from the `CONTRIBUTING.md`.
```
rannn505, line 86: why newline here?

py/configu/commands/eval_command.py
```diff
@@ -24,90 +24,92 @@ class EvaluatedConfigOrigin(Enum):
...
-    total=False,
-)
+class EvalCommandReturnResult(TypedDict):
+    origin: EvaluatedConfigOrigin
+    source: str
+    value: str
+
+
+class EvalCommandReturnValue(TypedDict):
+    context: EvalCommandReturnContext
```
rannn505, line 42: this should be aligned with the ts sdk the eval now returns a flat object that the export then transforms to the result

py/configu/commands/eval_command.py
```diff
@@ -135,16 +135,13 @@ def _eval_from_store_set(
                 value["result"]["value"] = store_config.value
         return result
 
-    def _eval_from_schema(
-        self, result: EvalCommandReturn
-    ) -> EvalCommandReturn:
+    @staticmethod
```
rannn505, line 138: why is it marked as staticmethod?

rannn505 (changes requested):
@RonConfigu great work overall here :), see minor comments and lets schedule for next week to talk on the HC vault store :)

py/configu/commands/eval_command.py
RonConfigu: semantics. it doesn't use `self` (JS:`this`). it's also marked "private" so even though it's static it's not for outside use.
if it bothers you we can lose it. it's just semantics.

py/README.md
RonConfigu: ide default. python project shouldn't have lines longer 79 - 119 chars. 
since it's not python, reverted to long line.

py/README.md
RonConfigu: ide default. python project shouldn't have lines longer 79 - 119 chars. 
since it's not python, reverted to long line.

py/DOCS.md
RonConfigu: it's the "index" for the auto-generated docs.
image

py/configu/commands/eval_command.py
RonConfigu: fixed. please notice on data3 EvalCommand, even though i'm passing configs overrides for `GREETING` it doesn't override it. this is because this key is from the previous and current schema. is this intended? 
image

py/configu/core/command.py
```diff
@@ -7,7 +7,7 @@
 
 
 @dataclass
-class Command(Generic[CommandReturn]):
+class Command(ABC, Generic[CommandReturn]):
```
jamesL92, line 10: I've been wondering about this for a while...it might be better to use `Protocol[CommandReturn]` for duck-typing here.

My thinking is:
- We don't need to have `parameters` in the base class, allowing for parameter-less commands.
- We don't need to import subclass `Command` in various places
- Users wishing to implement their own commands don't have to subclass `Command`, their class just needs to implement a `run` method.

We can retain runtime-checking with the (runtime_checkable)[https://docs.python.org/3/library/typing.html#typing.runtime_checkable] decorator if we want.

py/configu/core/config_schema.py
```diff
@@ -129,8 +114,7 @@ def parse(cls, scheme: "ConfigSchema") -> Dict[str, Cfgu]:
                     error_message(
                         f"invalid key {key}",
                         error_location + [key],
-                        f"path nodes mustn't contain "
-                        f"reserved words '${key}'",
+                        f"path nodes mustn't contain " f"reserved words '${key}'",
```
jamesL92, line 117: `" f"` can be removed

jamesL92 (changes requested):

jamesL92:
> * changed max-line-length to 88 for better readability
> * added DOCS.md for first page in auto-docs.
> * added some docs for example
> * changed api signature for `Commands` and `ConfiguConfigStore`
> 
> @jamesL92 If you have the time and will, take a look. I also fixed `gcp-store` You can use this command to generate the docs:
> 
> ```shell
> poetry run pdoc -o docs configu --logo https://app.configu.com/logo192.png --favicon https://app.configu.com/favicon.ico
> ```

Fantastic changes overall, in particular I love the new class signatures from accepting a dict parameter to named args. I've made a few suggestions based on things I've found, some are more personal style preferences so feel free to accept/reject those suggestions as you see fit :+1:

py/configu/core/command.py
jamesL92: if we do decide to go down the  `Protocol` route, there's probably some additional refactoring work needed to achieve that. `ABC` is already a vast improvement, so would suggest this is fine for this PR and we look at converting it at a later date.

py/configu/stores/json_file_store.py
```diff
@@ -17,60 +10,44 @@ class JsonFileConfigStore(ConfigStore):
...
         """
         Reads JSON file contents
-        :return: The file contents, represented as a list of key/set/value
+        :return: The file contents, represented as a list of Config
         dicts
         """
         with open(self._path, mode="r", encoding="utf-8") as json_file:
             data = json.load(json_file)
-        return config_store_contents_from_dict(data)
+        return [Config(**args) for args in data]
```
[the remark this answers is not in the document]
RonConfigu: Yes. doing my best to get rid of convertors from generated.py

py/configu/core/config_schema.py
```diff
@@ -139,8 +123,7 @@ def parse(cls, scheme: "ConfigSchema") -> Dict[str, Cfgu]:
                         "invalid type property",
                         error_location + [key, cfgu.type.value],
                     ),
-                    f"type '{cfgu.type.value}' must come with"
-                    f" a pattern property",
+                    f"type '{cfgu.type.value}' must come with" f" a pattern property",
```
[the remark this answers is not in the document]
RonConfigu: true. this is because of the new line width. thanks!

py/configu/core/config_schema.py
RonConfigu: true. this is because of the new line width. thanks!

py/configu/commands/eval_command.py
```diff
@@ -24,90 +23,92 @@ class EvaluatedConfigOrigin(Enum):
...
+    """
+
+
+class EvalCommandParameters(TypedDict):
+    store: ConfigStore
+    set: ConfigSet
+    schema: ConfigSchema
+    configs: Dict[str, str]
+    validate: bool
+    previous: EvalCommandReturn
```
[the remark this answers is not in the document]
RonConfigu: `previous` definitely should be `Optional`. We will only use it the user is using  `pipe notation` to pipe multiple `EvalCommand`s
 Nevertheless, please explain why defaulting to `None` makes a difference? or how would you make it optional without specifying default value?

py/configu/commands/eval_command.py
```diff
@@ -24,90 +23,92 @@ class EvaluatedConfigOrigin(Enum):
...
     parameters: EvalCommandParameters
 
-    def __init__(self, parameters: EvalCommandParameters) -> None:
+    def __init__(
+        self,
+        *,
+        store: ConfigStore,
+        set: ConfigSet,
+        schema: ConfigSchema,
+        configs: Dict[str, str] = None,
```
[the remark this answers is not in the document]
RonConfigu: Why do you think it's not optional anymore? I just removed the typing because it makes no difference.
It's bad practice as far as I know to init mutable params at __init__ (eg `configs={}`)
It wouldn't change the logic in anyway but it keeps the __init__ clean of values, hinting the user that

py/configu/commands/eval_command.py
```diff
@@ -24,90 +23,92 @@ class EvaluatedConfigOrigin(Enum):
...
-        "validate": Optional[bool],
-        "previous": Optional[EvalCommandReturn],
-    },
-    total=False,
-)
+
+class EvalCommandReturn(Dict[str, EvalCommandReturnValue]):
+    """
+    Dict of keys:EvalCommandReturnValue`
+    """
```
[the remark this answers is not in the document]
RonConfigu: I actually wen't back and forth here with this one..
I'll take you suggestion, It will also clear the code from all the `EvalCommandReturn()` calls

py/configu/commands/eval_command.py
```diff
@@ -24,90 +23,92 @@ class EvaluatedConfigOrigin(Enum):
...
+                previous=previous,
+            )
+        )
 
     def _eval_from_configs_override(
         self, result: EvalCommandReturn
     ) -> EvalCommandReturn:
         if not self.parameters.get("configs"):
-            return {}
+            return EvalCommandReturn()
```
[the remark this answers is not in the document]
RonConfigu: https://github.com/configu/configu/pull/163#discussion_r1285609026
yep

py/configu/stores/gcp_secret_manager.py
```diff
@@ -23,24 +28,22 @@ def get_by_key(self, key: str) -> str:
...
+        secret = Secret()
+        secret.replication.automatic = {}
+        try:
+            self._client.create_secret(
+                parent=f"projects/{self._project_id}",
+                secret_id=key,
+                secret=secret,
+            )
+        except (Exception,):
+            pass
```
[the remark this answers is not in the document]
RonConfigu: This exception only occur when `set()` is trying to create a created secret. if it fails (meaning there's a secret already) than just update it.
this logic repeats in other store. try: create / catch: update

py/configu/core/command.py
RonConfigu: I'd look into Protocol. for now let's keep it this way and as you suggested, PR Protocol.

RonConfigu:
Thanks @jamesL92 🙏 
Pushed with changes.

py/configu/commands/eval_command.py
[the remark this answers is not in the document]
jamesL92: Not optional in that the `Optional` typing was removed (see previous comment). I certainly didn't know that mutable init default params are bad practice...read up a little and found out why. I certainly learned something new today :)

Based on that I'd agree that `None` is the right default value to use here, however in that case I'd suggest changing the typing here to either `Optional[Dict[str,str]]` or `Union[Dict[str,str],None]`, so that type checking doesn't identify this as an error.

py/configu/stores/gcp_secret_manager.py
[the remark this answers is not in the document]
jamesL92: OK yeah that makes sense 👍 I somehow read this change without the context of the next line (`self._add_secret_version(formatted_key, value)`).

py/configu/commands/eval_command.py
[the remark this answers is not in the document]
RonConfigu: I'll take it, Optional is back.
Thanks

py/configu/commands/eval_command.py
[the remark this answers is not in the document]
RonConfigu: Superb. Optional is back since prev comment so we're good here.

jamesL92 (approved):

RichardAkman (approved):
LGTM, amazing work! @RonConfigu and thank you @jamesL92 for your insight!

py/configu/commands/eval_command.py
rannn505: @RonConfigu Im not sure #171 is a bug we should discuses it.

py/configu/commands/upsert_command.py
```diff
@@ -60,8 +67,7 @@ def run(self):
                     error_message(
                         f"invalid config key '{key}'",
                         scope_location,
-                        f"key '{key}' must be declared"
-                        f" on schema {schema.path}",
+                        f"key '{key}' must be declared" f" on schema {schema.path}",
```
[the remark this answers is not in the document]
rannn505: @jamesL92 great catch

py/configu/stores/json_file_store.py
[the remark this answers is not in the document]
rannn505: @RonConfigu is Config is part of the generated code?
`quicktype` is used to enforce the base interfaces of all the SDKs so I wonder, do you guys think it is irrelevant for python or do you use it in some other parts without the convertors?

rannn505:
@RonConfigu please merge master to this branch before you merge to main and release.
@jamesL92 Thank you very much for the help with the review, your comments are valuable and important! ❤️

rannn505 (approved):

py/configu/stores/json_file_store.py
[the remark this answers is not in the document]
jamesL92: yup, both versions are provided by the `quicktype` generated code. This is simpler however since it doesn't rely on an auxillary alias of `ConfigStoreContentsElement`, nor does it rely on a utility `config_store_contents_from_dict` function. In theory this takes us one step closer to entirely removing some of the unnecessary auxillary types from `quicktype` entirely.

py/configu/core/command.py
RonConfigu: opened
https://github.com/configu/configu/issues/199#issue-1850979943

Issue #160, opened 2023-07-09
Implement `Keyv` `ConfigStore` in Node.js SDK


gabrielcipriano:
**Description**
- Plan a `Config` accommodation in the `Keyv` storage.
- Add `KeyvConfigStore` class to the Node.js SDK stores.
- We do not need to add `keyv` as a dependency. We just need to respect the `keyv` interface.
- the client would pass the keyv instance as a parameter of the KeyvConfigStore class, favouring decoupling:
```typescript
const keyvMongo = new Keyv('mongodb://user:pass@localhost:27017/dbname');

const configStore = new KeyvConfigStore(keyvMongo)
```

**Motivation**
[Keyv](https://github.com/jaredwray/keyv) provides a consistent interface for key-value storage across multiple backends via storage adapters.

Official Storage Adapters
[etcd](https://github.com/jaredwray/keyv/tree/main/packages/etcd): Etcd storage adapter
[memcache](https://github.com/jaredwray/keyv/tree/main/packages/memcache): Memcache storage adapter
[mongo](https://github.com/jaredwray/keyv/tree/main/packages/mongo): MongoDB storage adapter
[mysql](https://github.com/jaredwray/keyv/tree/main/packages/mysql): MySQL/MariaDB storage adapter
[offline](https://github.com/jaredwray/keyv/tree/main/packages/offline): Offline storage adapter
[postgres](https://github.com/jaredwray/keyv/tree/main/packages/postgres): PostgreSQL storage adapter
[redis](https://github.com/jaredwray/keyv/tree/main/packages/redis): Redis storage adapter
[sqlite](https://github.com/jaredwray/keyv/tree/main/packages/sqlite): SQLite storage adapter
[tiered](https://github.com/jaredwray/keyv/tree/main/packages/tiered): Tiered storage adapter

> "The official storage adapters are covered by [over 150 integration tests](https://github.com/jaredwray/keyv/actions/workflows/tests.yaml) to guarantee consistent behaviour. They are lightweight, efficient wrappers over the DB clients making use of indexes and native TTLs where available."

As it's a stablished lib with many adapters ('backends'), I believe that it will be a great adition to the project.
That way we can take advantage of the already developed keyv adapters: Virtually, any keyv adapter would be a valid Configu store.

related issues:
#145 
#146

**Additional context**
[https://github.com/jaredwray/keyv](https://github.com/jaredwray/keyv)

gabrielcipriano:
[...]

Issue #146, opened 2023-07-03
Implement `MongoDB` `ConfigStore` in Node.js SDK


rannn505:
**Description**

- Plan a `Config` accommodation in the MongoDB storage.
- Add `MongoDBConfigStore` class to the Node.js SDK stores.
- We can use the native [mongodb](https://www.npmjs.com/package/mongodb) driver library during implementation.
- Or we can leverage the [ORMConfigStore](https://github.com/configu/configu/blob/main/ts/packages/node/src/stores/ORM.ts) implemented with `typeorm` that add a [mongodb support](https://typeorm.io/data-source-options#mongodb-data-source-options).

**Motivation**

Extending Configu's [available stores](https://configu.com/docs/config-store/#available-stores) collection.
This ConfigStore will be made available in the Configu CLI since the Node.js SDK is used internally.

**Additional context**

- https://www.mongodb.com/
- https://configu.com/docs/config-store/
- https://github.com/configu/configu/blob/main/types/ConfigStore.ts
- https://github.com/configu/configu/tree/main/ts/packages/node/src/stores

![image](https://github.com/configu/configu/assets/11806915/e603c3c3-703c-46d2-91a6-3c308089a8ce)

Rohxn16:
Imma take look here

SanjanaSogimatt:
Is the issue still open? Can I work on it?

rannn505:
@Rohxn16 , @SanjanaSogimatt still open, feel free to go for it <3

Tushar-4781:
Hi @rannn505 i want to give this issue a shot. Just wanted to confirm my understanding
1. we need to Create a class called MongoDBConfigStore that extends the base ConfigStore class and 
2. implement the get and set methods to interact with MongoDB for retrieving and storing key-value pairs.

rannn505:
@Tushar-4781 Indeed, you can spin up Configu with other DB based ConfigStore and see the results.

rannn505:
Hi @Tushar-4781,

You've nailed the understanding perfectly! Your task is indeed to create a MongoDBConfigStore class that extends our base ConfigStore class and to implement the get and set methods for MongoDB interactions.

Feel free to draw inspiration from the other DB stores within the Node SDK. Additionally, you have the freedom to opt for either the native MongoDB driver or the TypeORM based store - choose whichever aligns best with your implementation strategy.

We’re here to assist you throughout your contribution journey. Eagerly awaiting to review your work!

rannn505:
[...]

Issue #145, opened 2023-07-03
Implement `etcd` `ConfigStore` in Node.js SDK


rannn505:
**Description**

- Plan a `Config` accommodation in the etcd storage.
- Add `EtcdConfigStore` class to the Node.js SDK stores.
- A quick search indicates that [etcd3](https://www.npmjs.com/package/etcd3) library is the most "suitable" for use during implementation.
- We would be happy to hear more suggestions, or we can implement the APIs ourselves.

**Motivation**

Extending Configu's [available stores](https://configu.com/docs/config-store/#available-stores) collection.
This ConfigStore will be made available in the Configu CLI since the Node.js SDK is used internally.

**Additional context**

- https://etcd.io/
- https://etcd.io/docs/v3.3/integrations/
- https://configu.com/docs/config-store/
- https://github.com/configu/configu/blob/main/types/ConfigStore.ts
- https://github.com/configu/configu/tree/main/ts/packages/node/src/stores

![image](https://github.com/configu/configu/assets/11806915/e603c3c3-703c-46d2-91a6-3c308089a8ce)

PR #167, opened 2023-07-13
Implement `Keyv` `ConfigStore` in @configu/node

gabrielcipriano:
related issues: 
#160 
#145
#146 

### Description

This PR adds a new `KeyvConfigStore` class to the Node.js SDK stores.

* `KeyvConfigStore` allows users to store configuration data in a Keyv store.
* Keyv is a popular key-value storage library that supports a variety of backends, including MongoDB, MySQL, Redis, and SQLite.
* `KeyvConfigStore` is implemented using the `KeyValueConfigStore` base class.
* This ensures that `KeyvConfigStore` has the same API as the other ConfigStore implementations.

### Motivation

* Provide users with more flexibility in how they store configuration data.
* Keyv is a well-established library with a wide range of supported backends.
* By adding a `KeyvConfigStore` class to the Node.js SDK, we can give users the ability to store their configuration data in any Keyv-supported backend, such as mongoDB and Etcd 

### Implementation

* `KeyvConfigStore` is implemented as a simple wrapper around the `Keyv` library.
* The class takes a `keyvInstance` property, which is an instance of the `Keyv` library.
* `KeyvConfigStore` then uses the `keyvInstance` property to store and retrieve configuration data.

### Testing

TODO - where could we place unit tests for the config stores? - maybe open a issue on this?

### Documentation

TODO - where could I place the docs of this config store?

-----

* If there is any questions or suggestions to this PR, please feel free to say.


ts/packages/node/package.json
```diff
@@ -51,6 +51,7 @@
     "typeorm": "^0.3.16"
   },
   "devDependencies": {
-    "@types/lodash": "^4.14.195"
+    "@types/lodash": "^4.14.195",
+    "keyv": "^4.5.2"
```
rannn505, line 55: shouldn't `keyv` be under "dependencies" prop?

ts/packages/node/package.json
```diff
@@ -51,6 +51,7 @@
...
   }
-}
```
rannn505, line 56: not sure why is this diff, probably EOL?

ts/packages/node/src/stores/Keyv.ts
```diff
@@ -0,0 +1,40 @@
...
+ * import Keyv from 'keyv';
+ * import { KeyvConfigStore } from '@configu/ts-node';
+ * const keyvMongo = new Keyv('mongodb://user:pass@localhost:27017/dbname');
+ * const configStore = new KeyvConfigStore({ keyvInstance: keyvMongo });
+ * ```
+ */
+export class KeyvConfigStore extends KeyValueConfigStore {
+  client: Keyv;
+
+  constructor({ keyvInstance }: KeyvConfigStoreConfiguration) {
```
rannn505, line 23: if we wish to make this store available in the cli, we should somehow allow passing the Keyv constructor parameter type (e.g the mongo connection string / other store connection options) as the `KeyvConfigStoreConfiguration` so it can be supplied as strings and booleans  and call `new Keyv` in the constructor.

ts/packages/node/src/stores/Keyv.ts
```diff
@@ -0,0 +1,40 @@
+/* eslint-disable spellcheck/spell-checker */
+import { KeyValueConfigStore } from '@configu/ts';
+import type Keyv from 'keyv';
+
+export type KeyvConfigStoreConfiguration = { keyvInstance: Keyv };
+
+/**
```
rannn505, line 7: Nice work on the doc string 🥇

ts/packages/node/src/stores/Keyv.ts
```diff
@@ -0,0 +1,40 @@
...
+    return value ?? '';
+  }
+
+  protected async upsert(key: string, value: string): Promise<void> {
+    await this.client.set(key, value);
+  }
+
+  protected async delete(key: string): Promise<void> {
+    await this.client.delete(key);
+  }
```
rannn505, line 39: Smoothest implementation ever! 💯

rannn505 (changes requested):
@gabrielcipriano great work! please see me comments here, sorry for taking a bit long to review this, LMK if you need any clarification or assistance with finalizing the PR.

ts/packages/node/package.json
gabrielcipriano: oh,  probably hahah

ts/packages/node/src/stores/Keyv.ts
gabrielcipriano: Thanks! I thought it would be good to have a example popping out from the IDE while typing

ts/packages/node/package.json
gabrielcipriano: I've been thinking about this, and keyv could be a dependency of the project. But I'm not sure if the keyv backends could be. Once they are numerous, (counting only the official ones, there are 9), we would be adding a lot of dependencies to the project.

Based on my understanding of the documentation, - "You can choose the most suitable ConfigStore based on your specific needs and workflow" - users of ConfigStores will generally use only one implementation. 

For example, someone who only uses the AWS Secrets Manager ConfigStore is unlikely to use any of the other implementations. Therefore, there is no need to have mysql, mssql, pg, axios, etc. as dependencies of the project. If they need another implementation in addition to the AWS Secrets Manager, let's say Postgres, they would download the Configu package with only the Postgres implementation, to use alongside the Configu core package.

Of course, what I'm suggesting here would be a drastic change, and it needs to be discussed and studied to see if it makes sense for the project as a whole.

But what I propose for now is trying to do something similar to this in the `KeyvConfigStore`: we could add only the Keyv as a dependency of the project, and if the client wants to use a custom Keyv Backend, they would have to install the adapter in addition to the Configu package. 

That would be something similar to what keyv suggests in their documentation:

>Usage
Install Keyv.
```bash
npm install --save keyv
```
>By default everything is stored in memory, you can optionally also install a storage adapter.
```bash
npm install --save @keyv/redis
npm install --save @keyv/etcd
```
>Create a new Keyv instance, passing your connection string if applicable. Keyv will automatically load the correct storage adapter.

```ts
const Keyv = require('keyv');

// One of the following
const keyv = new Keyv();
const keyv = new Keyv('redis://user:pass@localhost:6379');
const keyv = new Keyv('etcd://localhost:2379');
```

ts/packages/node/src/stores/Keyv.ts
gabrielcipriano: That's a great idea! 

We could write our constructor based on the Keyv constructor parameter (e.g the mongo connection string / other options), so that way we would instantiate the keyv instance

but for the mongo connection string to work, the client would have to install the mongo adapter alongside the Configu package. So we would need to make it clear on the docs.

To use KeyvConfigStore with mongodb, It could be something similar to this:
### Install
```bash
npm install @configu/node @keyv/mongo
```
### Use
```ts
const configStore = new KeyvConfigStore('mongodb://user:pass@localhost:27017/dbname')
```
in the example above Keyv will automatically load the correct storage adapter, as the client installed @keyv/mongo alongside @configu/node

gabrielcipriano:
Hey @rannn505, firstly thanks a lot for your feedback! 

I believe I managed to express some questions I had, and as soon as we have something defined, I will proceed with the work on the PR.

ts/packages/node/package.json
rannn505: @gabrielcipriano,
You've brought up a compelling perspective. Balancing flexibility with simplicity is always a delicate act, and the idea of modularizing ConfigStores is intriguing.

I see the merits in your suggestion. By allowing users to install only the specific stores they need, we could reduce unnecessary bloat in their projects. However, this could become challenging with SDKs in languages where package management isn't as flexible as npm. Additionally, there's the consideration that a significant portion of users might leverage multiple stores concurrently.

I believe the best approach might be to engage the broader community in this discussion. Your idea deserves wider attention, so I propose opening a dedicated issue for it and inviting feedback. Concurrently, I'll investigate potential solutions that allow us to package each language SDK without mandating all dependencies. I'll link the issue back here once it's live.

Thank you for prompting this discussion—it's insights like these that help refine and improve our project!

cc @configu/dev

ts/packages/node/package.json
rannn505: I'm curious about enforcing a new line at EOF. Which OS are you using? I'm wondering why Prettier isn't catching this, or if there's another tool we might need to incorporate.

ts/packages/node/package.json
rannn505: For now, before merging it Im suggesting that we will move `keyv` to the `dependencies` section

PR #172, opened 2023-07-24
Implement LaunchDarkly ConfigStore

RonConfigu:


RichardAkman:
Please connect the PR to the related issue 🙏

RonConfigu:
- Update lib in saas and update.
- after release create new branch for CLI

ts/packages/node/src/stores/LaunchDarkly.ts
```diff
@@ -0,0 +1,173 @@
+import { Config, ConfigStore, ConfigStoreQuery } from '@configu/ts';
+import axios, { Axios } from 'axios';
+import assert from 'node:assert';
```
RichardAkman, line 3: I reviewed all usages of `assert` in this store implementation and I honestly couldn't find any justification for its usage. Unless you can clarify why you decided to use it, I highly suggest keeping store implementations as lean and as clear as possible, especially considering this is the only instance of `assert` being used in the entire codebase.

ts/packages/node/src/stores/LaunchDarkly.ts
```diff
@@ -0,0 +1,173 @@
...
+    throw new Error('No valid queries');
+  }
+
+  private assertNotSettingToRoot(configs: Config[]) {
+    assert.strictEqual(
+      configs.some((config) => config.set === ''),
+      false,
+      'ConfigSet cannot be empty string',
+    );
+  }
```
RichardAkman, line 145: I don't see any reason why this should be its own method if it's only used by the `set` method.

RichardAkman, line 145: These methods should be merged into a single method that validates sets. Any invalid set should cause the operation to throw. The validation method shouldn't return anything, it should only validate.

Is there a reason why in `validateQueries` you don't validate and throw if one of the sets is the root set?

ts/packages/node/src/stores/LaunchDarkly.ts
```diff
@@ -0,0 +1,173 @@
...
+import axios, { Axios } from 'axios';
+import assert from 'node:assert';
+import * as fs from 'fs';
+
+export type LaunchDarklyConfigStoreConfigurations = { apitoken: string; defaultproject: string } | string;
+
+export class LaunchDarklyConfigStore extends ConfigStore {
+  private client: Axios;
+  private readonly projectKey: string;
+  private environments: string[];
```
RichardAkman, line 11: I don't see a reason to hold the environment in memory if we refetch them every time the class instance is used. If there are helper methods that need the environments, let's pass them to them using the latest environments that we anyway fetch at the start of every `get` or `set`.

ts/packages/node/src/stores/LaunchDarkly.ts
```diff
@@ -0,0 +1,173 @@
...
+      return this.getEnvFeatureFlags(env, keys);
+    } catch (e) {
+      return [];
+    }
+  }
+
+  async set(configs: Config[]): Promise<void> {
+    this.assertNotSettingToRoot(configs);
+    await this.updateEnvironments();
+    configs.forEach(this.upsert.bind(this));
```
RichardAkman, line 171: Is there a reason why `promise.all` isn't used here?

ts/packages/node/src/stores/LaunchDarkly.ts
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  private async getFeatureFlag(config: Config) {
+    const { data: featureFlag } = await this.client.get(`/flags/${this.projectKey}/${config.key}`);
+    return featureFlag;
+  }
+
+  private async createEnvironment(config: Config) {
+    await this.client.post(`/projects/${this.projectKey}/environments`, {
+      color: 'FFFFFF',
```
RichardAkman, line 49: For the sake of clarity, let's add a comment that explains this field is required for env creation

ts/packages/node/src/stores/LaunchDarkly.ts
```diff
@@ -0,0 +1,173 @@
...
+      headers: {
+        Authorization: configs.apitoken,
+      },
+      responseType: 'json',
+    });
+    this.environments = [];
+  }
+
+  private async patchUpdate(config: Config, patchData: Record<string, any>) {
+    const { data } = await this.client.patch(`/flags/${this.projectKey}/${config.key}`, patchData, {
```
RichardAkman, line 31: Please add a comment above each REST API call with a link to the LaunchDarkly docs that explains the endpoint

ts/packages/node/src/stores/LaunchDarkly.ts
```diff
@@ -0,0 +1,173 @@
...
+    return data.variations.find((variation: any) => variation.value === config.value);
+  }
+
+  private async updateDefaultFallthroughVariation(config: Config, variationId: string) {
+    const updateDefaultVariationData = {
+      instructions: [
+        {
+          kind: 'updateDefaultVariation',
+          OnVariationValue: config.value,
+          OffVariationValue: '',
```
RichardAkman, line 84: Why is `OffVariationValue` needed here?

ts/packages/node/src/stores/LaunchDarkly.ts
```diff
@@ -0,0 +1,173 @@
...
+    const updateFallthroughVariationData = {
+      environmentKey: config.set,
+      instructions: [
+        {
+          kind: 'updateFallthroughVariationOrRollout',
+          variationId,
+        },
+      ],
+    };
+    await this.patchUpdate(config, updateDefaultVariationData);
```
RichardAkman, line 97: I'm not sold on why we should update the default value.

ts/packages/node/src/stores/LaunchDarkly.ts
```diff
@@ -0,0 +1,173 @@
...
+    });
+    this.environments = [...this.environments, config.set];
+  }
+
+  private async createFeatureFlag(config: Config) {
+    const createData = {
+      clientSideAvailability: {
+        usingEnvironmentId: true,
+        usingMobileKey: true,
+      },
```
RichardAkman, line 61: Why did you pick these values and not the defaults? Also, if there are defaults, do we even need to provide these?
image

ts/packages/node/src/stores/LaunchDarkly.ts
```diff
@@ -0,0 +1,173 @@
...
+      key: config.key,
+      name: config.key,
+      variations: [{ value: config.value }, { value: '' }],
```
RichardAkman, line 64: The empty string variant is very arbitrary, I'm not sure if this is the right approach. If we do this then we can never create a boolean or any other type of feature flag by using configu:
image

We need to consider if this makes sense for whoever uses this store.

ts/packages/node/src/stores/LaunchDarkly.ts
```diff
@@ -0,0 +1,173 @@
...
+    return featureFlag;
+  }
+
+  private async createVariation(config: Config) {
+    const createVariationData = {
+      instructions: [{ kind: 'addVariation', value: config.value }],
+    };
+    const data = await this.patchUpdate(config, createVariationData);
+    return data.variations.find((variation: any) => variation.value === config.value);
+  }
```
RichardAkman, line 76: This will only work if the feature flag is of type string. If the user has existing flags that are not of type string, this will fail with the following error:
image

We need to rethink this logic, we can't simply make this assumption because it will either make this store unusable or force users to recreate their feature flags with the string type

ts/packages/node/src/stores/LaunchDarkly.ts
```diff
@@ -0,0 +1,173 @@
...
+export class LaunchDarklyConfigStore extends ConfigStore {
+  private client: Axios;
+  private readonly projectKey: string;
+  private environments: string[];
+
+  constructor(configurations: LaunchDarklyConfigStoreConfigurations) {
+    super('launch-darkly');
+    const configs =
+      typeof configurations === 'string'
+        ? JSON.parse(fs.readFileSync('ldc.json').toString())[configurations]
```
RichardAkman, line 17: The path shouldn't be hardcoded, what if they renamed the file to something else?

ts/packages/node/src/stores/LaunchDarkly.ts
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  async get(queries: ConfigStoreQuery[]): Promise<Config[]> {
+    try {
+      const [env, keys] = await this.validateQueries(queries);
+      return this.getEnvFeatureFlags(env, keys);
+    } catch (e) {
+      return [];
+    }
+  }
```
RichardAkman, line 166: Why are we returning an empty array? As far as I'm aware, other stores return an empty value on keys that we failed to fetch. Please explain so I can understand the rationale.

RonConfigu: so if something goes wrong you'd prefer an exception?

ts/packages/node/src/stores/LaunchDarkly.ts
```diff
@@ -0,0 +1,173 @@
...
+        (await this.createVariation(config));
+      if (existingVariation) featureFlag = await this.updateDefaultFallthroughVariation(config, existingVariation._id);
+    } catch (e) {
+      if (e.response.status === 404) featureFlag = await this.createFeatureFlag(config);
+      else throw e;
+    }
+    if (featureFlag && !featureFlag.environments[config.set].on) {
+      const onData = {
+        environmentKey: config.set,
+        instructions: [{ kind: 'turnFlagOn' }],
```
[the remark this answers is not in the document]
RonConfigu: This is the core principle of feature-flag. we should probably go through this in-person so you'd understand the reason for this kind of choices Peleg made.

ts/packages/node/src/stores/LaunchDarkly.ts
RonConfigu: they can't and shouldn't.
this is how FeatureFlag's sdk expecting the file to named.

ts/packages/node/src/stores/LaunchDarkly.ts
RonConfigu: I'm "pretty sure" line 130 filters root-set from the queries.

They way EvalCommand composing these queries is hierarchical. So it's certain that this method will receive query for root-set.
If it's not clear i'll show you in-person in the code.
This is also the reason these are separated.

I will change the name for `validateQueries` to whatever you'd say but it will always return the set and the keys.

BTW, the only reason to use `if` and `throw` instead of `assert` here is TS giving me errors that don't exist.

ts/packages/node/src/stores/LaunchDarkly.ts
RonConfigu: I mean... you're generally right.. but feels like exaggeration. it's an array of strings. in JS i think it's as memory-optimized as plain string since it's also an array of chars.
But if it's crucial to you i'll make it `getEnv..` and it will return them instead.

ts/packages/node/src/stores/LaunchDarkly.ts
RonConfigu: Maybe i'm missing something but it exactly the same... 
i'll use promise.all

how about this?
```javascript
const upsertPromises = configs.map(this.upsert.bind(this));
await Promise.all(upsertPromises);
```

if you'd decide that I shouldn't keep the envs in mem then i'll have to create them all first.

ts/packages/node/src/stores/LaunchDarkly.ts
RonConfigu: sure

ts/packages/node/src/stores/LaunchDarkly.ts
RonConfigu: sure

ts/packages/node/src/stores/LaunchDarkly.ts
RonConfigu: we're not supporting any other types then strings. we will convert the string to CFGU types on our end.

ts/packages/node/src/stores/LaunchDarkly.ts
RonConfigu: actually, I think you're right. at first i thought having the default is they way to activate this value but then i found `updateDefaultFallthroughVariation` so i think this is enough.
i'll do some tests if that correct i'll remove it.

RonConfigu: You're correct. actually, more than correct. good catch on that. removed the default part.

ts/packages/node/src/stores/LaunchDarkly.ts
RonConfigu: No.
We'll never give the user the option to control these settings from our SDK so we'll turn them on when creating.
He can then change the way they deliver via the UI as he will for any other feature-key

ts/packages/node/src/stores/LaunchDarkly.ts
RonConfigu: We always use strings in their DB and will take care of the type from our end via CFGU

ts/packages/node/src/stores/LaunchDarkly.ts
RonConfigu: That's a good point.. I think i can solve this with JSON.parse/stringfy on our end

ts/packages/node/src/stores/LaunchDarkly.ts
RonConfigu: Please to use try/catch

ts/packages/node/src/stores/LaunchDarkly.ts
RonConfigu: Please remove from Class

ts/packages/node/src/stores/LaunchDarkly.ts
RichardAkman: Add a comment that links to an explanation of this file.

ts/packages/node/src/stores/LaunchDarkly.ts
RonConfigu: change to defaults

ts/packages/node/src/stores/LaunchDarkly.ts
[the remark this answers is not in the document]
RonConfigu: Please don't turn flag on

ts/packages/node/src/stores/LaunchDarkly.ts
```diff
@@ -0,0 +1,154 @@
...
+    if (queries.some((query) => query.set.split('/').length > 1))
+      throw new Error('Subsets are not supported in LaunchDarklyConfigStore');
+    const environments = await this.getEnvironments();
+    const noRootQueries = queries.filter((query) => query.set !== '' && environments.includes(query.set));
+    if (noRootQueries[0]) {
+      const env = noRootQueries[0].set;
+      const keys = noRootQueries.map((value) => value.key);
+      return this.getEnvFeatureFlags(env, keys);
+    }
+    throw new Error('No valid queries');
```
RonConfigu, line 128: When no values found. should this return empty array or something else?

ts/packages/node/src/stores/LaunchDarkly.ts
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  async set(configs: Config[]): Promise<void> {
+    if (configs.some((config) => config.set === '')) throw new Error('ConfigSet cannot be empty string');
+    const environments = await this.getEnvironments();
+    const upsertPromises = configs.map(async (config) => {
```
RichardAkman, line 134: NIT: how about we use `setConfigsPromises` to align the naming with other stores like `KeyValue` store

ts/packages/node/src/stores/LaunchDarkly.ts
```diff
@@ -0,0 +1,154 @@
...
+
+  private async getEnvFeatureFlags(env: string, keys: string[]): Promise<Config[]> {
+    const { data: featureFlags } = await this.client.get(`/flags/${this.projectKey}?env=${env}`);
+    return featureFlags.items
+      .filter((featureFlag: any) => keys.includes(featureFlag.key))
+      .map((featureFlag: any) => {
+        const onValue = Object.entries(featureFlag.environments[env]._summary.variations)
+          .filter((value: [string, any]) => value[1].isFallthrough)
+          .map((value) => featureFlag.variations[parseInt(value[0], 10)].value)[0];
+        return { key: featureFlag.key, set: env, value: JSON.stringify(onValue) };
```
RichardAkman, line 113: `JSON.stringify(onValue)` is problematic if the value is already of type script

ts/packages/node/src/stores/LaunchDarkly.ts
RichardAkman: Should always return an empty array if no values are found, I also suggest we apply the following change to this method, tell me what you think:

```node.js
    const getConfigsPromises = _(queries)
      .filter((query) => query.set !== '' && environments.includes(query.set))
      .groupBy('set')
      .map((groupedConfigs) => {
        if (groupedConfigs[0]) {
          return this.getEnvFeatureFlags(groupedConfigs[0].set, _.map(groupedConfigs, 'key'));
        }
        return null;
      })
      .compact()
      .value();

    const getConfigsResults = await Promise.all(getConfigsPromises);
    const storedConfigs = _.flatten(getConfigsResults);

    return storedConfigs;
```

ts/packages/node/src/stores/LaunchDarkly.ts
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  private async getEnvFeatureFlags(env: string, keys: string[]): Promise<Config[]> {
+    const { data: featureFlags } = await this.client.get(`/flags/${this.projectKey}?env=${env}`);
+    return featureFlags.items
+      .filter((featureFlag: any) => keys.includes(featureFlag.key))
+      .map((featureFlag: any) => {
+        const onValue = Object.entries(featureFlag.environments[env]._summary.variations)
+          .filter((value: [string, any]) => value[1].isFallthrough)
+          .map((value) => featureFlag.variations[parseInt(value[0], 10)].value)[0];
```
RichardAkman, line 112: Why are we using `_summary` instead of `fallthrough`:
image

ts/packages/node/src/stores/LaunchDarkly.ts
```diff
@@ -0,0 +1,154 @@
...
+    };
+    return this.patchUpdate(config, updateFallthroughVariationData);
+  }
+
+  private async getEnvFeatureFlags(env: string, keys: string[]): Promise<Config[]> {
+    const { data: featureFlags } = await this.client.get(`/flags/${this.projectKey}?env=${env}`);
+    return featureFlags.items
+      .filter((featureFlag: any) => keys.includes(featureFlag.key))
+      .map((featureFlag: any) => {
+        const onValue = Object.entries(featureFlag.environments[env]._summary.variations)
```
RichardAkman, line 110: `onValue` is misleading, we're taking the fallthrough value no?

ts/packages/node/src/stores/LaunchDarkly.ts
RichardAkman: There are still missing comments with links, please add them above each rest API call

ts/packages/node/src/stores/LaunchDarkly.ts
RonConfigu: cool. done.
I'm sure you tested this change on every scenario before suggesting it.

ts/packages/node/src/stores/LaunchDarkly.ts
RonConfigu: Done.
Please make a note of it for future stores. `get()` should always return empty array if no exception raised.

ts/packages/node/src/stores/LaunchDarkly.ts
RonConfigu: Sure. but the english is incorrect. should be `setConfigPromises` we have many promises each set one config.

ts/packages/node/src/stores/LaunchDarkly.ts
```diff
@@ -0,0 +1,154 @@
...
+    this.client = axios.create({
+      baseURL: `${configs.server ? configs.server : 'https://app.launchdarkly.com'}/api/v2`,
+      headers: {
+        Authorization: configs.apitoken,
+      },
+      responseType: 'json',
+    });
+  }
+
+  // https://apidocs.launchdarkly.com/tag/Feature-flags#operation/patchFeatureFlag
```
RichardAkman, line 28:
```suggestion
  // * https://apidocs.launchdarkly.com/tag/Feature-flags#operation/patchFeatureFlag
```

ts/packages/node/src/stores/LaunchDarkly.ts
RonConfigu: I believe you meant `string`? take a look at the new code

ts/packages/node/src/stores/LaunchDarkly.ts
RonConfigu: sure. changed across the file.

ts/packages/node/src/stores/LaunchDarkly.ts
RonConfigu: please update the project. the links are there......

ts/packages/node/src/stores/LaunchDarkly.ts
```diff
@@ -0,0 +1,176 @@
...
+    super('launch-darkly');
+    // * ldc.json is the default config file name. see https://github.com/launchdarkly-labs/ldc
+    const fileConfigs =
+      configurations.ldcJsonPath && configurations.ldcConfiguration
+        ? JSON.parse(fs.readFileSync(configurations.ldcJsonPath).toString())[configurations.ldcConfiguration]
+        : {};
+    const configs = {
+      ...{ apitoken: configurations.apitoken, defaultproject: configurations.apitoken },
+      ...fileConfigs,
+    };
```
RichardAkman, line 27: `defaultproject: configurations.apitoken` is definitely a mistake.

I suggest we do this:
```node.js
    const { ldcJsonPath, ldcConfigurationKey, ...restConfigurations } = configurations;
    const ldcConfigurations =
      ldcJsonPath && ldcConfigurationKey
        ? JSON.parse(fs.readFileSync(ldcJsonPath).toString())[ldcConfigurationKey]
        : {};
    const configs = _.merge(restConfigurations, ldcConfigurations);
```

ts/packages/node/src/stores/LaunchDarkly.ts
```diff
@@ -0,0 +1,176 @@
...
+  ldcConfiguration?: string;
+}
+
+export class LaunchDarklyConfigStore extends ConfigStore {
+  private client: Axios;
+  private readonly projectKey: string;
+
+  constructor(configurations: LaunchDarklyConfigStoreParams) {
+    super('launch-darkly');
+    // * ldc.json is the default config file name. see https://github.com/launchdarkly-labs/ldc
```
RichardAkman, line 19: Remove the comment from here, let's move it up to the `ldcJsonPath`

ts/packages/node/src/stores/LaunchDarkly.ts
```diff
@@ -0,0 +1,176 @@
+import { Config, ConfigStore, ConfigStoreQuery } from '@configu/ts';
+import axios, { Axios } from 'axios';
+import * as fs from 'fs';
+import _ from 'lodash';
+
+export interface LaunchDarklyConfigStoreParams {
+  apitoken?: string;
+  defaultproject?: string;
+  ldcJsonPath?: string;
+  ldcConfiguration?: string;
```
RichardAkman, line 10:
```node.js
  // * https://github.com/launchdarkly-labs/ldc
  ldcJsonPath?: string;
  ldcConfigurationKey?: string;
}
```

RonConfigu: sure

ts/packages/node/src/stores/LaunchDarkly.ts
RonConfigu: definitely. thanks.
i see you've grown in to lodash way of life.

RichardAkman (approved):
Amazing work!! 🌟 LGTM 🌟

Issue #135, opened 2023-06-29
Go SDK implementation!


jamesL92:
**Description**
An initial go implementation of the SDK. Maybe only supporting in-memory/json/configu stores initially for simplicity.

**Motivation**
Golang is increasing in popularity, would be nice to support it. 
Also I wouldn't mind trying to come up with an initial implementation myself (for a bit of practice), just not sure if Configu is ready to support/maintain a go library yet - checking here first so that I'm not wasting my time :sweat_smile:

**Additional context**

rannn505:
Dear @jamesL92,

Firstly, thank you for showing such interest in contributing to Configu. We highly appreciate your initiative and are excited about the idea of introducing a Go implementation of the SDK. Your suggestions and enthusiasm are valuable to the growth and expansion of the project.

In terms of readiness, Configu is indeed prepared to support and maintain a Go library. We believe in the adaptability and expansiveness of our project, and introducing a new language implementation would be a great step forward. The fact that GoLang is rising in popularity, as you've pointed out, makes it an excellent choice.

Your offer to contribute an initial implementation is more than welcome. Contributions like these are what make open-source projects like Configu thrive. For sure, it won't be a waste of your time.

I suggest we continue this conversation both here and through email/zoom meeting so we can dive deeper into the details of the SDK skeleton implementation and deployment process. Feel free to reach out to me at ran@configu.com. We are eager to support you on this journey.

Looking forward to our collaboration and once again, thank you for your willingness to contribute.

Best Regards,
Ran and the whole Configu team.

rannn505:
@jamesL92  couple of resources I gathered as preparation for the implementation:

### `quicktype` npm script to the add to the package.json
```json
  scripts: {
    "prepare:go": "npm run types -- --lang go [--package configu ?] --out go/.../generated.go"
}
```
You can test the different outputs with all the specific go flags here: https://app.quicktype.io/#l=go
Also you can see the actual flags name running `npx quicktype --help` in the root this:
![image](https://github.com/configu/configu/assets/11806915/3f5447bb-69f4-43a4-bceb-dc3180447cae)

### equivalent dependencies to other SDKs
[ts deps](https://github.com/configu/configu/blob/main/ts/packages/ts/package.json#L33) | [py deps](https://github.com/configu/configu/blob/main/py/pyproject.toml#L31)

#### HTTP Client
- ts: [Axios](https://axios-http.com/docs/intro)
[...]

PR #173, opened 2023-07-25
Go SDK implementation

jamesL92:
Initial implementation of Golang implementation, per #135 .


jamesL92:
I've fixed a number of the issues I was having with this SDK, all now seems to be working as expected.

Manually tested an upsert & eval command with JsonFileStore, which all seems to be working as expected (at least when compared to the python SDK).

Things left to do that I can think of before this PR is ready to go:
- Switch package name from `configu` to `github.com/configu/configu` in `go.mod` and related areas.
- Documentation (docstrings and the likes)
- Populate examples in `examples` dir
- Add a README for the go package
- CI pipeline
- Overall project boilerplating (add `prepare:go` command to top-level `package.json` and so on)
- (Possibly) look into restructuring the SDK so that everything can be imported from top-level (rather than core/commands/stores separately)

if anyone can think of anything else to add in then by all means let me know :+1:

jamesL92:
Done pretty much everything I can think is necessary for this PR, so should be ready for review.

Two notes:
Tempted to handle documentation in a separate PR, since there's a little bit of extra complexity in terms of deploying/updating documentation. I think the general approach should be to rely on (pkg.go.dev)[https://pkg.go.dev/about#adding-a-package] for this, as it's a pretty standard location for golang documentation. Will likely also need to build this into a CD pipeline though to update docs when necessary (but can be done manually until that's done).

Likewise, I've omitted any kind of CD pipeline for this as with golang there's no need to release to a package manager (such as pypi or npmjs) - github essentially acts as the package manager for you (i.e. via `go get github.com/configu/configu/go`) with versioning handled by git tags. :+1:

examples/go-sdk/main.go
```diff
@@ -0,0 +1,28 @@
...
+	"fmt"
+
+	configu "github.com/configu/configu/go"
+)
+
+func main() {
+	store := configu.InMemoryStore{}
+	set, _ := configu.NewConfigSet("test")
+	schema, _ := configu.NewConfigSchema("get-started.cfgu.json")
+	configu.UpsertCommand{
```
gabrielcipriano, line 13: clever interface 👌🚀

gabrielcipriano (commented):
great work @jamesL92! my code review was more learning from you than anything else

go/configu_store.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+	defer resp.Body.Close()
+	var responseBody []Config
+	json.NewDecoder(resp.Body).Decode(&responseBody)
+	return responseBody
+}
+
+func (s ConfiguConfigStore) Set(configs []Config) {
+	payload, _ := json.Marshal(setRequestPayload{configs})
+	request, _ := http.NewRequest("PUT", "https://api.configu.com/config", bytes.NewBuffer(payload))
```
sashaLutsker, line 42: `json.Marshal` and `http.NewRequest` return an error. Best practice to check errors.

go/configu_store.go
```diff
@@ -0,0 +1,51 @@
...
+	request.Header.Set("Content-Type", "application/json")
+	request.Header.Set("Org", s.Org)
+	request.Header.Set("Authorization", fmt.Sprintf("Bearer %s", s.Token))
+	(&http.Client{}).Do(request)
```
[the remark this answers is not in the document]
jamesL92: Good suggestion, I've updated error checking throughout so any errors will bubble up through the commands.

go/configu_store.go
jamesL92: :+1: good suggestion. I've changed so that any errors here are bubbled up.

go/README.md
```diff
@@ -0,0 +1,67 @@
+# @configu/py
```
pelegpor, line 1:
```suggestion
# @configu/go
```

.github/configs/go-release-drafter.yml
```diff
@@ -0,0 +1,27 @@
+include-paths:
+  - "go/"
+name-template: "go-v$RESOLVED_VERSION"
```
rannn505, line 3: @jamesL92 In reference to your comment about versioning with GitHub, will these tags serve as the versions? Also, could you clarify how a user can pin a specific version? Thanks!

examples/go-sdk/main.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"fmt"
+
+	configu "github.com/configu/configu/go"
+)
+
+func main() {
+	store := configu.InMemoryStore{}
+	set, err := configu.NewConfigSet("test")
```
rannn505, line 11: @jamesL92 For the instance creation, is the New prefix necessary? If it isn't, I'd recommend renaming it to simply ConfigSet, ConfigSchema, and so on.

go/README.md
```diff
@@ -0,0 +1,67 @@
...
+}
+```
+
+## Reference
+
+[//]: # Link here once we have a reference page.
+
+## Contributing
+
+### Requirements
```
rannn505, line 51: This is awesome README thanks!!

go/config_schema.go
```diff
@@ -0,0 +1,313 @@
+package configu
+
+import (
+	"fmt"
+	"os"
+	osPath "path"
+	"regexp"
+	"strings"
+
+	"github.com/go-playground/validator/v10"
```
rannn505, line 10: 🥇

go/config_store.go
```diff
@@ -0,0 +1,7 @@
+package configu
+
+type IConfigStore interface {
+	Get(queries []ConfigStoreQuery) ([]Config, error)
+	Set(configs []Config) error
+	GetType() string
```
rannn505, line 6: @jamesL92 Should we consider adding this method across all SDKs? Or os it essentially serves the same purpose as a public property in Go?

package.json
```diff
@@ -13,7 +13,8 @@
     "types": "quicktype --src-lang typescript --src ./types/*.ts",
     "prepare:ts": "npm run types -- --lang typescript --out ts/packages/ts/src/types/generated.ts --prefer-unions",
     "prepare:vscode": "npm run types -- --lang schema --out ts/packages/vscode/public/cfgu.schema.json && json -I -f ts/packages/vscode/public/cfgu.schema.json -e 'this.$ref=\"#/definitions/ConfigSchemaContents\"' -e 'this.$schema=undefined'",
-    "prepare:py": "npm run types -- --lang python --python-version 3.7 --nice-property-names --out py/configu/core/generated.py"
+    "prepare:py": "npm run types -- --lang python --python-version 3.7 --nice-property-names --out py/configu/core/generated.py",
+    "prepare:go": "npm run types -- --lang go --package configu --out go/generated.go"
```
rannn505, line 17: @jamesL92 Do you believe any specific flags are necessary? you can run `npx quicktype --help` to view the options.

package.json
```diff
@@ -23,4 +24,4 @@
     "lint-staged": "^13.2.2",
     "quicktype": "^23.0.48"
   }
-}
+}
```
rannn505, line 27: I'm curious about the cause of this. Which IDE and OS are you using? https://github.com/configu/configu/pull/167#discussion_r1268777475

rannn505 (approved):
@jamesL92 You are the best! 🚀

package.json
jamesL92: VSCode with Linux (Ubuntu) :+1:

go/config_store.go
jamesL92: Essentially the equivalent of a public property in Go. There's also a public `Type` variable in the generated types [here](https://github.com/jamesL92/configu/blob/feat/go-implementation/go/generated.go#L214), but couldn't find a "clean" way of making use of it for concrete config stores (since the concept of subtyping doesn't translate well to Go). Interfaces are typically the idiomatic way to do this sort of thing in Golang.

For most OO languages (python/TS currently, but in future things like Java/C#) there's not really any benefit in adding this, although I think Rust will need a similar thing when that SDK is created/implemented.

examples/go-sdk/main.go
jamesL92: We _can_ do without it, however there are tradeoffs to consider. It's even possible with the current implementation.

This is typically the way to handle custom constructor logic in Go - e.g. for `ConfigSet`s verifying that the path is valid and for `ConfigSchema`s ensuring that the config schema uses the correct extension. The other way would be to have this logic in a separate method:
```go
set := configu.ConfigSet{"test"}
err := set.ValidatePath()
```

I would recommend sticking with the `NewConfigSet` semantics, since:
- it's more in line (logically) with other SDKs
- the user doesn't have to "remember" to call a separate method for such custom logic.

Happy to go with either approach, let me know if you'd like me to change this and I'll do so accordingly :+1:

.github/configs/go-release-drafter.yml
rannn505: @jamesL92 I think the README of the go dir is a good place for now, we are working on a dedicated docs site for the OSS alone, when it be ready I will let you know.

Issue #174, opened 2023-07-31
Implement `AWSParameterStore` `ConfigStore` in Node.js SDK 


idanela:
**Description**

- Plan a `Config` accommodation in AWS Parameter Store.
- Add `AWSParameterStoreConfigStore` class to the Node.js SDK stores.
- We can use the [AWS SSM Client](https://www.npmjs.com/package/@aws-sdk/client-ssm) Node.js SDK during implementation.

**Motivation**

Extending Configu's [available stores](https://configu.com/docs/config-store/#available-stores) collection.
This ConfigStore will be made available in the Configu CLI since the Node.js SDK is used internally.

**Additional context**

- https://docs.aws.amazon.com/systems-manager/latest/userguide/systems-manager-parameter-store.html
- https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-ssm/
- https://configu.com/docs/config-store/
- https://github.com/configu/configu/blob/main/types/ConfigStore.ts
- https://github.com/configu/configu/tree/main/ts/packages/node/src/stores

PR #175, opened 2023-07-31
Implement AWS Parameter Store

idanela:
Implementation of AWS Parameter Store #174


ts/packages/node/src/stores/AWSParameterStore.ts
```diff
@@ -0,0 +1,38 @@
...
+  // * https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/ssm/command/GetParameterCommand/
+  async getByKey(key: string): Promise<string> {
+    const command = new GetParameterCommand({ Name: key });
+    const res = await this.client.send(command);
+    return res?.Parameter?.Value ?? '';
+  }
+
+  // * https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/ssm/command/PutParameterCommand/
+  async upsert(key: string, value: string): Promise<void> {
+    const command = new PutParameterCommand({ Name: key, Value: value, Type: ParameterType.STRING, Overwrite: true });
```
jamesL92, line 29: I would recommend `ParameterType.SECURE_STRING` here - since a big portion of Configu's use case entails storing things like DB connections/passwords/other highly sensitive data, a lot of users will require these to be encrypted at rest.

Going down this route means you'll need to add `WithDecryption: true` to your `GetParameterCommand` too.

RichardAkman (approved):
Amazing work and a very sleek implementation 🌟 

As a follow-up to this awesome contribution to our Node.js SDK, we will add the store to our CLI which will allow the CLI users to enjoy the fruits of your labor too 🤩

To do this, we will update the `@configu/lib` package as such:
- Update the `@configu/node`  dependency version to the newest version
- Update the `src/integrations/store.ts` file to include the new store by:
  - Add `aws-parameter-store` to StoreType
  - Add `'aws-parameter-store': 'AWS Parameter Store'` to `STORE_LABEL`
  - Add `'aws-parameter-store': 'https://aws.amazon.com/systems-manager/features/#Parameter_Store'` to `STORE_WEBSITE`

After we release a new version of `@configu/lib` we will then update the `@configu/cli` package to support AWS Parameter Store as such:
- Update the `@configu/node`  dependency version to the newest version
- Update the `@configu/lib`  dependency version to the newest version
- Update the `src/helpders/store.ts` file by adding `'aws-parameter-store': AWSParameterStoreConfigStore` to `ConfigStoreCtor`

Thanks again and we can't wait to see further contributions 🥇

PR #178, opened 2023-08-01
Add AWS Parameter Store to @configu/lib integrations

RichardAkman:


ts/packages/lib/src/integrations/store.ts
```diff
@@ -43,6 +45,8 @@ export const STORE_WEBSITE: Record<StoreType, string> = {
   configu: 'https://configu.com/',
   'json-file': 'https://www.json.org/json-en.html',
   'hashicorp-vault': 'https://www.vaultproject.io/',
+  'aws-parameter-store':
```
rannn505, line 48: change to the main product page as in secret-manager please

rannn505 (approved):

ts/packages/lib/src/integrations/store.ts
RichardAkman: There is no main product page as it's a feature of AWS Systems Manager. There is [this section](https://aws.amazon.com/systems-manager/features/#Parameter_Store) on the product page but it doesn't really link to anything useful in the context of the parameter store. What would you prefer?

ts/packages/lib/src/integrations/store.ts
```diff
@@ -43,6 +45,8 @@ export const STORE_WEBSITE: Record<StoreType, string> = {
...
+    'https://docs.aws.amazon.com/systems-manager/latest/userguide/systems-manager-parameter-store.html',
```
rannn505, line 49: Use this https://aws.amazon.com/systems-manager/features/#Parameter_Store

Issue #182, opened 2023-08-06
Fix `KeyValue` `ConfigStore` set method in TS SDK


RichardAkman:
**Actual behavior**

The `set` method of the `KeyValue` ConfigStore within the TS SDK currently does not properly implement the upsert paradigm like other stores because any existing values that are not part of the upsert within a set get deleted. Additionally, the delete method never gets called leaving garbage empty sets.

**To Reproduce**

- Use any ConfigStore that extends KeyValue such as AWS Parameter Store and upsert to an existing set with values that aren't part of the upsert. The result will be that only the upserted values will remain and the other values are discarded.
- Use Upsert to delete all values within any ConfigStore that extends KeyValue such as AWS Parameter Store and see that the set remains without any values.

**Expected behavior**

Existing values that are not explicitly meant for deletion should remain and any set that is left without values after upsert should be deleted from the config store.

PR #183, opened 2023-08-06
Fix `KeyValue` `ConfigStore` set method

RichardAkman:
To resolve issue #182, The set method will now fetch any values that are contained within a set and attempt to merge them if the set is a stringified JSON and override the value if it's not. Additionally, if after the merge/override, the remaining value is an empty object or an empty string, we will attempt to delete the key to conserve resources, something that didn't work previously.


RonConfigu (approved):
lgtm

rannn505:
@RichardAkman Change the name of the issue and PR to something that implies what is the real fix here!

PR #216, opened 2023-08-17
Move `Keyv` to dependencies from dev dependencies

RichardAkman:
Following a recent discussion on correctly placing the `keyv` library in the `@configu/node` package.json, the placement should definitely be in the `dependencies` and not `devDependencies`.


RonConfigu (approved):
👍

PR #220, opened 2023-08-17
Catch-up python CFGU types with ts

jamesL92:
Additional CFGU types have been added to ts, add them to python too. Where `pyvalidator` doesn't have the appropriate validator methods to match javascript's `validator` packages, ensure the same regex/lists are used.


RonConfigu:
@jamesL92 Thanks for the super quick action 🚀 🥇 

Goes without saying, Great job and many thanks!

Even though I can't promise we'll always use validator.js/py as long as we are, seems a little messy having the iso file added to our repo, especially when it's only for python.

Instead, I've opened a PR with py-validator for mime and iso6391 (https://github.com/theteladras/py.validator/pull/20) and hopefully they're reactive and we could just update the package. if not we'll discuss other options.

RonConfigu:
Hi @jamesL92 

https://pypi.org/project/py.validator/ is now update with is_mime_type and is_iso6391
Please update the package and use these.

Thanks!

RonConfigu (approved):
Perfect 🥇 

Thanks!

PR #222, opened 2023-08-17
Implement fallback exception when cfgu-type doesn't have a validator function

RonConfigu:


RonConfigu:
@jamesL92 

Please review the suggested changes and comment as needed.
You can also DM me @ discord i'm `Markeitect`

Thanks!
Ron

py/configu/commands/eval_command.py
```diff
@@ -158,24 +158,36 @@ def _validate_result(self, result: EvalCommandReturn):
...
-                if not type_test(*test_values):
-                    raise ValueError(
+                    if not type_test(*test_values):
+                        raise ValueError(
+                            error_message(
+                                f"invalid value type for key '{key}'", error_scope
+                            ),
+                            f"value '{test_values[0]}' must be a " f"'{cfgu.type}'",
+                        )
+                except KeyError:
```
jamesL92, line 178: I'd recommend wrapping only line 162 in the `try-except`. That way it's clear what caused the error to happen.
For example, if in future a poorly constructed `type_test` ends up raising a `KeyError` we end up raising a misleading error message - the particular `CfguType` is _technically_ supported, however there is a bug in the particular implementation.

py/configu/commands/eval_command.py
```diff
@@ -158,24 +158,36 @@ def _validate_result(self, result: EvalCommandReturn):
...
                         ),
-                        f"value '{test_values[0]}' must be a " f"'{cfgu.type}'",
+                        f"type '{cfgu.type.value}' is not yet supported in this SDK. "
+                        f"For the time being, please utilize the String type. "
+                        f"We'd greatly appreciate it if you could open an issue "
+                        f"regarding this at "
+                        f"https://github.com/configu/configu/issues/new/choose "
+                        f"so we can address it in future updates.",
                     )
+
```
jamesL92, line 190: In cases such as these I tend to recommend the following pattern:
```py
try:
    <<potentially raising code>>    
except KeyError as e:
    raise <<new exception>> from e
```
That way the underlying stacktrace is persisted and the user gets more detailed error messaging, making it easier to debug.

Additionally, do all lines in the error message have to be f-strings? If not we should probably only make the first line an f-string (not sure if that's supported by python syntax :thinking:).

py/configu/commands/upsert_command.py
```diff
@@ -66,37 +66,49 @@ def run(self):
...
+                    if cfgu.type == CfguType.REG_EX:
+                        of_type += f"({cfgu.pattern})"
+                    raise ValueError(
+                        error_message(
+                            f"invalid config value '{value}' for key '{key}'",
+                            error_scope,
+                            f"value '{value}' must be of type '{of_type}'",
+                        )
                     )
+            except KeyError:
```
jamesL92, line 102: Same as above, we can probably significantly reduce the amount of logic we're wrapping in the try block here.

py/configu/commands/upsert_command.py
```diff
@@ -66,37 +66,49 @@ def run(self):
...
+                raise KeyError(
```
jamesL92, line 103: `raise KeyError(...) from e` pattern. Also removal of extra f-strings if possible.

py/configu/core/config_schema.py
```diff
@@ -130,28 +130,40 @@ def parse(cls, scheme: "ConfigSchema") -> Dict[str, Cfgu]:
...
+                        if not type_test(*test_values):
+                            raise ValueError(
+                                error_message(
+                                    "invalid default property",
+                                    error_scope + [key, "default"],
+                                ),
+                                f"{cfgu.default} must be of type {cfgu.type.value}"
+                                f" or match Regex",
+                            )
+                    except KeyError:
```
jamesL92, line 155: As in previous comments.

jamesL92 (changes requested):

py/configu/commands/eval_command.py
ShriekinNinja: Agreed. 
Fixed.
Thanks!

py/configu/commands/eval_command.py
ShriekinNinja: Syntax supported.
Fixed.
Thanks

py/configu/commands/upsert_command.py
ShriekinNinja: done

py/configu/commands/upsert_command.py
ShriekinNinja: done

py/configu/core/config_schema.py
ShriekinNinja: done

RichardAkman (approved):
LGTM ⭐ 🚀

PR #225, opened 2023-08-19
General chores and fixes

rannn505:
- update ts deps
- remove configu-store from node and use only on ts
- fix login command for other envs then dev and prod
- move CONFIGU_* env vars logic to cli
- fix prettier EOL from auto to lf
- add the official configu-icon to the repo
- create initial cd-go
- add go sdk to the readme table
- detach oss assets from configu app
- lib populated ConfigSchema collection


.github/workflows/cd-go.yml
```diff
@@ -0,0 +1,73 @@
...
+
+      # - uses: aws-actions/configure-aws-credentials@v2
+      #   with:
+      #     aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      #     aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      #     aws-region: ${{ secrets.AWS_REGION }}
+
+      # ----------
+      # src: https://go.dev/doc/modules/publishing, https://pkg.go.dev/about#adding-a-package
+      - name: Bump module version and Publish to golang proxy
```
rannn505, line 36: @jamesL92 happy for your input on this pipeline. should I add any steps?

Issue #229, opened 2023-08-24
Ensure the go CD pipeline handles tags of the format go/vX.X.X


jamesL92:
**Description**
Currently the CD pipeline expects tags to be of the format `go-vX.X.X`. However in order for the `go` ecosystem to correctly identify and work with semantic versioning, semantic release tags should be of the format `go/vX.X.X`.

**Motivation**
So that `go get` can correctly identify and compare semantic versions, for stability/upgrades/etc...

**Additional context**
https://go.dev/doc/modules/managing-source#multiple-module-source details how to handle versioning of a multi-module repo. Although the example is mainly intended for repos with multiple go modules, the key similarity here is how to handle module versioning in a subdirectory of a repo.

I've done some testing on this repo: https://github.com/jamesL92/versiontest to verify what the correct behaviour should be. See that it correctly identifies tagged versions here: https://pkg.go.dev/github.com/jamesL92/versiontest/go
<img width="167" alt="image" src="https://github.com/configu/configu/assets/8472199/3443188c-aa09-47b3-8abf-46d5a88d2c03">

PR #230, opened 2023-08-24
(chore) Support go releases with go/vX.X.X tags

jamesL92:
#229


.github/actions/deploy-start-action/action.yml
```diff
@@ -51,6 +51,10 @@ runs:
           channel=next
         else
           echo "couldn't resolve release channel based on the pushed tag"
+          echo "for go releases:"
```
rannn505, line 54: tbd: @rannn505 migrate all cd-*.yml to work with the <package>/v<X.Y.Z> convention and enforce it via gh rule.

rannn505 (approved):
🚀

Issue #212, opened 2023-08-16
Unit tests across all SDKs


jamesL92:
**Description**
All SDKs here currently have no/next-to-no meaningful unit tests. These should be added across all supported SDKs

**Motivation**
- All the usual reasons for unit testing, too many to list here
- User confidence in adopting Configu's SDKs
- Having (essentially) the same unit tests across SDKs enables us to ensure consistency across different SDKs

**Additional context**
Unit test support is already added in the CI pipelines :rocket:

jamesL92:
@rannn505 is it better to keep this issue open until we also have unit testing in place for the Go SDK (not sure how well tested the TS sdks are, and whether they also need tests added)

PR #232, opened 2023-08-25
Add python unit tests

jamesL92:
This addresses #212 for the Python SDK.


py/tests/core/test_config_schema.py
```diff
@@ -0,0 +1,77 @@
...
+    "tests/staticfiles/config_schemas/invalid/default_and_template.cfgu.json",
+    "tests/staticfiles/config_schemas/invalid/nonexistent_type.cfgu.json",
+    "tests/staticfiles/config_schemas/invalid/invalid_default.cfgu.json",
+))
+def test_invalid_config_schemas(path):
+    with pytest.raises(ValueError):
+        ConfigSchema.parse(ConfigSchema(path))
+
+@pytest.mark.parametrize("path", (
+    # "tests/staticfiles/config_schemas/invalid/bad_extension.json", // TODO: This currrent fails. Is this a bug?
```
jamesL92, line 26: @RonConfigu @rannn505 Wasn't sure if this was a bug or not. As far as I was aware we should be throwing a `ValueError` if trying to use a config schema without the extension `.cfgu.<<ext>>`.
If so we should probably raise that as a bug, and I'll keep this test commented out for now.
If not, I'll remove this comment :+1:

py/tests/core/test_config_schema.py
```diff
@@ -0,0 +1,77 @@
...
+  ("IPv4", "192.168.0.0", True),
+  ("IPv4", "not-an-ipv4", False),
+  ("IPv6", "0000:0000:0000:0000:0000:0000:0000:0001", True),
+  ("IPv6", "not-an-ipv6", False),
+  ("Domain", "example.com", True),
+  ("Domain", "not-a-domain", False),
+  ("URL", "https://example.com", True),
+  ("URL", "not-a-url", False),
+  ("ConnectionString", "user@service/path?opts", True),
+  # ("ConnectionString", "", False), # TODO: Couldn't find any string that doesn't match the regex.
```
jamesL92, line 60: @RonConfigu any ideas on this? As far as I can tell it'll match literally any string, in which case the regex may be too broad? Alternatively if you know of a string that doesn't match the regex please let me know and I'll use it for this test.

py/tests/commands/test_export_command.py
```diff
@@ -0,0 +1,72 @@
...
+    eval_output = EvalCommand(
+        store=store,
+        set=config_set,
+        schema=schema,
+        configs={"GREETING": "hello"}
+    ).run()
+    ExportCommand(eval_output).run()
+    assert os.environ.get("GREETING") == "hello"
+
+#TODO: This test seems to be failing. Is this a bug in the export command or intended behaviour?
```
jamesL92, line 59: @RonConfigu / @rannn505 Am I right in thinking that the intended behaviour of `override=False` is to only set environment variables if the _don't_ already exist (i.e. not overriding existing env vars)?

jamesL92:
Would have liked to unit test a lot of the other stores as well, although these get significantly trickier.

For `AWSSecretsManagerConfigStore` we could use the [moto](https://docs.getmoto.org/en/latest/docs/services/secretsmanager.html) library, but not sure if this exists for other providers.

Another route for most of them could be to mock the `requests` library, big downside to this is that the tests lose all their value in so doing (can't test that configs are round-tripped correctly). I very much want to avoid taking this approach.

I'd very much welcome any suggestions anyone has around this, but for the time being I'd rather save myself that headache and worry about it in a separate PR :sweat_smile:

py/tests/core/test_config_schema.py
RonConfigu: Correct 💯
Unit tests justifying their purpose, once again.
This will fix it:
config_schema.ph - line 69
```python
EXT = " | ".join(["".join(ext) for ext in zip([CFGU.EXT], TYPES.keys())])
```
removed `cycle` and added `[]`to `CFGU.EXT`. There was a time when I though CFGU.EXT will be a list and this is a leftover. Thanks for catching 🥇 

You could solve this in a different way if you choose.

py/tests/commands/test_export_command.py
RonConfigu: Correct 💯 
export_command.py line 42 should be:
```python
if override or key not in os.environ:
```
added `not` to the if statement

py/tests/core/test_config_schema.py
RonConfigu: We'll wait for Ran with this one. The question is in the RegEx itself and it maybe exactly what he want it to be.
Regardless. Good catch! 👍

RonConfigu:
Superb initiative and PR @jamesL92 🥇 Thank you!
The one with the RegEx is still open but great catches on the other two. I left suggestions for fixes.

I think your gut feeling is correct. You've done an amazing work here and we should keep it focused on the core sdk for now in my opinion as well.
Later PRs will cover extended features (ie. stores)
@rannn505 thoughts?

py/tests/core/test_config_schema.py
jamesL92: :+1: In that case I'll leave it commented out with a TODO comment for now, to be tackled once a decision has been made regarding the regex.

RonConfigu: Opened issue regarding this.
https://github.com/configu/configu/issues/233
Thanks @jamesL92

RonConfigu (approved):
Beautiful work @jamesL92 💯 
Thank you for your continuous contributions. Much appreciated 🥇

PR #242, opened 2023-08-28
🎨 Implementing MongoDB ObjectID to Cfgu Type

AceAbhi147:
Changes includes:

- Added the MongoId type in Cfgu type
- Introduced the ENUM for mongoID in generated files for go, py & ts files
- Enforced validation on the newly added field in the validators file for go, py & ts files.


rannn505 (approved):
@AceAbhi147, fantastic job! Thanks for such a swift implementation. :)

Issue #240, opened 2023-08-28
Implement `options` Property for Cfgu Declaration


rannn505:
**Description**
Introduce an `options` property to the Cfgu declaration. Represented as an array of strings, this property would restrict the value to be one from a predefined set. Specifying `options: ['X', 'Y', 'Z']` would be equivalent to using `type: "RegEx"` alongside `pattern: "^(X|Y|Z)$"`.

```json
{
  ...
  "CONFIG_KEY": {
    "type": "String",
    "options": ["X", "Y", "Z"]
  }
  ...
} 
```

**Motivation**
The aim is to bolster the flexibility and user-friendliness of the Cfgu declaration by introducing intuitive alternatives to using the `RegEx` and `pattern` properties.

**Additional context**

jamesL92:
Please assign this one to me. :+1: 

A couple of thoughts I've had around validation options here:
- Eval/Upsert commands should naturally check that the evaluated/upserted value should be in the schema `options`, if present.
- ConfigSchema parsing should have a couple of additional validation rules when `options` is present:
  - `options` shouldn't be an empty list
  - each option should be valid based on the type, e.g. a `Number` type can't have `foo` as its option.
  - if `default` is also present, the default value should be present in options.
  - `template` should not be present.

Please let me know if the above validation suggestions make sense, if there are any that should be changed/remove or if there are any additional validation rules that should be added.

rannn505:
@jamesL92, it's all yours! It really seems like you've thought of everything. Go ahead – I'm looking forward to it. :)

jamesL92:
One extra thing I just thought of - the options list shouldn't contain an empty string :sweat_smile:

rannn505:
@jamesL92, while you're working on it, if you come across any missing validations on other properties, such as 'depends' which is also a string array, please let me know. I'll create a bug for it. Thanks! :)

PR #243, opened 2023-08-28
Feat/cfgu options support

jamesL92:
Addresses #240

### DIFF
--- a/py/configu/commands/eval_command.py
+++ b/py/configu/commands/eval_command.py
@@ -185,7 +185,18 @@ class EvalCommand(Command[EvalCommandReturn]):
                         error_message(
                             f"invalid value type for key '{key}'", error_scope
                         ),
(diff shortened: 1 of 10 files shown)


rannn505 (approved):
Hi @jamesL92, excellent work on this feature! I'm planning to fine-tune it a bit further, but it's already in great shape. Let's also keep in mind to align the Go SDK with these updates in the future.

jamesL92:
Sure thing - I'll look into adding this support for the other SDKs. Would you prefer I add the support for TS/Go SDKs in this same PR, or have one PR per SDK?

rannn505:
Hi @jamesL92, from looking at the PR it seems like you implemented it on TypeScript and Python! It looks like we only have the Go SDK left to align with these changes. I suggest we proceed by merging this current pull request. Whenever you're ready, feel free to open a new one for the Go SDK updates. Keep up the great work! 😊

PR #250, opened 2023-09-07
Fix Eval command prioritizing default values over previous eval results

RichardAkman:


ts/packages/ts/src/commands/EvalCommand.ts
```diff
@@ -142,7 +142,10 @@ export class EvalCommand extends Command<EvalCommandReturn> {
         if (
           !mergedResult ||
           (mergedResult.result.origin === EvaluatedConfigOrigin.EmptyValue &&
-            current.result.origin !== EvaluatedConfigOrigin.EmptyValue)
+            current.result.origin !== EvaluatedConfigOrigin.EmptyValue) ||
+          (mergedResult.result.origin === EvaluatedConfigOrigin.SchemaDefault &&
+            current.result.origin !== EvaluatedConfigOrigin.EmptyValue &&
+            current.result.origin !== EvaluatedConfigOrigin.SchemaDefault)
```
rannn505, line 148: I think that if there are 2 schemas with default for the same `key` it should be the latest in the parmas to determine it so this condition at :148 can change this na?

RichardAkman: You are absolutely correct. In the case of two schemas with defaults for the same key, the default of the latest schema should take precedence during evaluation. That's the logic I aimed to implement with this change. The condition at line 148 is precisely what we need to make this logic work because we iterate over the result from right to left, making the condition appear 'reversed.' I have retested this, and it functions as expected. Please feel free to review it as well.

ts/packages/ts/src/commands/EvalCommand.ts
```diff
@@ -127,6 +127,28 @@ export class EvalCommand extends Command<EvalCommandReturn> {
     });
   }
 
+  private shouldOverrideOrigin(nextOrigin: EvaluatedConfigOrigin, previousOrigin?: EvaluatedConfigOrigin): boolean {
+    switch (previousOrigin) {
+      case undefined:
+        return true;
+      case EvaluatedConfigOrigin.EmptyValue:
+        return nextOrigin !== EvaluatedConfigOrigin.EmptyValue;
```
RonConfigu, line 135: Considering
- `const mergedIsEmpty = !mergedResult || mergedResult.result.origin === EvaluatedConfigOrigin.EmptyValue;`
- `if (mergedIsEmpty || shouldOverrideMerged) {`
This test is either irrelevant or not doing it's job.. if the previous is EmptyValue the override will happen no matter the result from here

Also, NIT: why not combine both tests to one helper method ?

RonConfigu (commented):
Please see comments 🙏

RonConfigu (approved):
Superb 💯

PR #255, opened 2023-09-19
Refactor `ConfigSchema`

rannn505:
- refactor ConfigSchema constructor parameters to support programmatic only api for browsers etc.
~~- add ConfigSchema static fromFile method~~
- ConfigError class instead of ERR util
- add Cfgu.template validation at ConfigSchema
- add REGEX and JSON_SCHEMA utils


rannn505:
@jamesL92, would you mind reviewing this change? I'd love to get your perspective on whether we can apply it with the Python and Go SDKs.

ts/packages/ts/src/ConfigSchema.ts
```diff
@@ -1,463 +1,496 @@
...
+    const reason = 'invalid options property';
+    if (_.isEmpty(cfgu.options)) {
+      throw new ConfigError(reason, `options mustn't be empty if set`);
     }
-    this.type = fileExt as ConfigSchemaType;
+    if (cfgu.template) {
+      throw new ConfigError(reason, `options mustn't set together with template properties`);
+    }
+    cfgu.options.forEach((option, idx) => {
+      // todo: rethink on this restriction - what if the user sees empty string as a valid value?
```
jamesL92, line 347: Fair point - the main concern was (as far as I remember) that the `UpsertCommand` treats upserting an empty string as deleting a key, which conflicts with empty strings being valid values. Worth re-thinking for sure, but may not be as simple as just allowing empty strings in `options`.

ts/packages/ts/src/types/types.test.ts
```diff
@@ -17,6 +17,22 @@ describe(`types`, () => {
...
+      };
+      expect(res).toThrow();
+    });
+    it(`throw from Convert.cfguToJson`, async () => {
+      const res = () => {
+        Convert.cfguToJson(errored as any);
+      };
+      expect(res).toThrow();
+    });
+  });
```
jamesL92, line 34: I must admit personally I don't _really_ see the value in testing autogenerated code directly like this - essentially this is duplicating tests that are (or should be) part of the `quicktype` package. I think it's better to assume these work as follows and test things that use this autogenerated code instead. 
Sort of follows the "test public code interfaces, not private ones"

That said, there's also no harm in having them here, so it's more a matter of taste

ts/packages/ts/src/utils.ts
```diff
@@ -1,24 +1,56 @@
...
-  message: string,
-  { location = [], suggestion = '' }: { location?: string[]; suggestion?: string } = {},
-) => {
-  return `${message}${!_.isEmpty(location) ? ` at ${location.join('.')}` : ''}${suggestion ? `, ${suggestion}` : ''}`;
-};
+export class ConfigError extends Error {
+  constructor(
+    public reason: string,
+    public hint: string = '', // code?: string,
+    public scope: [string, string][] = [],
```
jamesL92, line 9: This is fine I think, but **_only_** because a golang version hasn't been tagged yet. 

Why? Because in the Golang version this ConfigError is already a `struct` (closes golang has to a class) and therefore things like `message`, `location` and `suggestion` are already publicly referenced by name. The cautionary tale here is to try to limit potential breaking changes (such as renaming things) as much as possible in OSS projects. Frequent breaking changes to key parts of the codebase are likely to put people off actually using the library, due to the pain of having to change their code every time they want to update dependencies. Ideally we should be aiming to make dependency version updates as seamless as possible.

I very much support making ConfigError a class here (and the equivalent in python), however I'd probably still keep reason/hint/scope named as message/suggestion/location. If you really do want to rename these, I'd suggest phasing it out over multiple versions, with a deprecation warning in between.

ts/packages/ts/src/utils.ts
```diff
@@ -1,24 +1,56 @@
...
+
+  setHint(hint: string) {
+    this.hint = hint;
+    this.setMessage();
+  }
+
+  appendScope(scope: [string, string][]) {
+    this.scope = [...scope, ...this.scope];
+    this.setMessage();
+  }
```
jamesL92, line 38: Really not a fan of having mutable error classes like this. For one it makes it more difficult to trace the of the error to the lowest level it was raised, secondly it doesn't play nicely at all with error reporting tools like [Sentry](https://sentry.io/welcome/). Generally better than mutating an error is to create a new error from the existing one.

types/ConfigSchema.ts
```diff
@@ -1,15 +1,13 @@
 import { Cfgu } from './Cfgu';
 
-export type ConfigSchemaType = "json";
+export interface ConfigSchemaContentsValue extends Cfgu {};
+export type ConfigSchemaContents = { [ConfigKey: string]: ConfigSchemaContentsValue };
```
jamesL92, line 4: I wonder if this can be simplified a little - something like:
```suggestion
export type ConfigSchemaContents = Record<string, Cfgu>;
```
Ideally not even having an auxillary type and typing `ConfigSchema.contents` directly as `Record<string, Cfgu>` would  be nice, but I'm not sure what to do then about `ConfigSchema.parse` in the `ts` sdk.

jamesL92 (changes requested):
Aside from the mutable errors it looks OK to me.

I don't see any reason something similar couldn't be done for the Python/Golang SDKs. 👍

ts/packages/ts/src/ConfigSchema.ts
```diff
@@ -1,463 +1,496 @@
...
-  }
+    const schemaName = splittedPath.pop();
+    if (!schemaName) {
+      error.setHint(`name "${schemaName}" mustn't be empty`);
+      throw error;
+    }
+    if (!NAME(schemaName)) {
+      error.setHint(`name "${schemaName}" mustn't contain reserved words`);
+      throw error;
+    }
```
RichardAkman, line 488: This should be extracted and used by the constructor too, as it stands, schema names are only validated when loaded from a file and not by the constructor. It should be consistent.

rannn505:
@jamesL92 @RichardAkman, thank you both for your insightful code review. I am grateful for the feedback. I will address each of your comments and suggest new changes based on them on Tuesday morning. 
**Your expertise is truly invaluable in enhancing the quality of this project.**

ts/packages/ts/src/ConfigSchema.ts
rannn505: @RichardAkman and there the bug here with the path having `/` chars

ts/packages/ts/src/ConfigSchema.ts
rannn505: Totally get your point, @jamesL92. I've left a comment in the code to keep your thoughts in mind for later. Thanks a lot for highlighting this!

ts/packages/ts/src/types/types.test.ts
rannn505: I appreciate your perspective on testing autogenerated code. While I see the merit in focusing on public code interfaces, I opted to include basic error tests for the methods we actively use as a safeguard against potential issues with quicktype, such as unexpected bugs or breaking changes. Even though I'm keeping them for now, I fully agree that our project's test coverage can be improved and it's high time to develop more substantial tests.

ts/packages/ts/src/utils.ts
rannn505: @jamesL92, your caution regarding breaking changes and the importance of seamless updates is absolutely valid. I share the sentiment that we must be judicious with modifications that impact our users directly. Even though we're in a 0.X.Y version and the project is relatively young, which might afford us some flexibility with changes, I do acknowledge that each alteration requires thorough consideration. While this PR does suggest a significant shift for the ConfigSchema API, it’s a delicate balance to manage evolution while ensuring stability. Your suggestion of phased updates and deprecation warnings is a prudent approach that we should embed in our development process moving forward. I'll certainly bear this in mind with regard to the naming within ConfigError and our overall strategy for rolling out updates.

For now I suggest we take the "move fast and break things" for us to get to the first stable 1.X.Y version

ts/packages/ts/src/utils.ts
rannn505: ABSOLUTELY!

types/ConfigSchema.ts
rannn505: I've pondered that as well. When initially integrating quicktype into the project, I created numerous tests. However, attempting to optimize any code in the types led to unsatisfactory generated results. Feel free to experiment with it and run `npm i` at the root dit afterward to observe the effects.

image

image

PR #257, opened 2023-09-21
Create fallback error message for not-implemented Cfgu types 

### LINKED ISSUE #221: Create fallback error message for not-implemented Cfgu types
**Description**

When trying to validate schema (EvalCommand/UpsertCommand/ConfigSchema.parse/etc..) if there is no validation function for the CfguType a informative message should be raised.

**Motivation**

**Additional context**

Sadaf-A:
This PR creates fallback error message for not-implemented `Cfgu types` in TS
Resolves #221

### DIFF
--- a/ts/packages/ts/src/commands/EvalCommand.ts
+++ b/ts/packages/ts/src/commands/EvalCommand.ts
@@ -1,7 +1,7 @@
 import _ from 'lodash';
 import { Command } from '../Command';
 import { Cfgu } from '../types';
-import { ERR, TMPL } from '../utils';
+import { ERR, TMPL, isStringInCfguType } from '../utils';
 import { ConfigStore } from '../ConfigStore';
 import { ConfigSet } from '../ConfigSet';
 import { ConfigSchema } from '../ConfigSchema';
@@ -101,6 +101,11 @@ export class EvalCommand extends Command<EvalCommandReturn> {
       const { context } = current;
       const { cfgu } = context;
 
+      if (!isStringInCfguType(current.context.schema)) {
+        throw new Error(
+          `Type '${current.context.schema}' is not yet supported in this SDK.\nFor the time being, please utilize the String type.\nWe'd greatly appreciate it if you could open an issue regarding this at https://github.com/configu/configu/issues/new/choose so we can address it in future updates.`,
+        );
+      }
       if (cfgu.template) {
         return {
           ...current,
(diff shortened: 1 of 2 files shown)


ts/packages/ts/src/utils.ts
```diff
@@ -30,3 +31,46 @@ export const TMPL = {
...
+  'SHA',
+  'SemVer',
+  'String',
+  'URL',
+  'UUID',
+]);
+
+export const isStringInCfguType = (input: string): boolean => {
+  return cfguTypeSet.has(input as CfguType);
+};
```
jamesL92, line 76: Rather than creating a new set, it's better to refer to the `VAL_TYPE` map [here](https://github.com/configu/configu/blob/main/ts/packages/ts/src/ConfigSchema.ts#L21). That way there's a single source of truth which `CfguType`s are supported, as well as the standard being "we support a CfguType if and only if we have a validator hooked up for it".

ts/packages/ts/src/commands/EvalCommand.ts
```diff
@@ -101,6 +101,11 @@ export class EvalCommand extends Command<EvalCommandReturn> {
       const { context } = current;
       const { cfgu } = context;
 
+      if (!isStringInCfguType(current.context.schema)) {
+        throw new Error(
+          `Type '${current.context.schema}' is not yet supported in this SDK.\nFor the time being, please utilize the String type.\nWe'd greatly appreciate it if you could open an issue regarding this at https://github.com/configu/configu/issues/new/choose so we can address it in future updates.`,
+        );
+      }
```
jamesL92, line 108: Better handled in `ConfigSchema.parse` rather than `EvalCommand`, since this would also apply to other commands like `UpsertCommand`.

jamesL92 (changes requested):
This is better handled in the `ConfigSchema` refactoring found in #255 , so may be worth closing this PR and relying on that one instead.

Sadaf-A:
@jamesL92 Thanks, I'll close the PR

rannn505:
@jamesL92, I truly appreciate your insightful comments – they shed light on important aspects of the project. 
@Sadaf-A, thank you for giving it a try. There's no definitive right or wrong approach here, both methods hold merit. However, I have a slight preference towards keeping the `utils` centered around handling runtime dependencies, and housing these kinds of validations within a centralized common place like the ConfigSchema class. This setup fosters clarity and consistency across our codebase. Your engagement is invaluable and it's through such discussions that we refine our approaches for the better.

PR #260, opened 2023-09-27
Fix CLI partially upserting config values containing the `=` character

### LINKED ISSUE #259: Value that contains the equal sign will ignore everything after the equal sign
**Actual behavior**
A value that contains the equal sign will ignore everything after the equal sign

**To Reproduce**
<!-- Steps to reproduce the behavior:
1. Upsarting config with a value that contains an equal sign 

**Expected behavior**
The value should contain the actual value without ignoring the equal sign and everything after

tal-testim:
Resolves #259


RichardAkman (approved):
That is a great catch and an even smoother fix! Thanks a lot for your first contribution! 🌟 🌟 🌟

PR #261, opened 2023-09-29
Fix HashiCorpVaultConfigStore upsert method not supporting string values

### LINKED ISSUE #256: Fix `HashiCorpVaultConfigStore` upsert method not supporting string values
**Actual behavior**

Hashicorp vault does not support storing string data in secrets, this means that in any instance where we upsert values as strings instead of JSONs, an error will be thrown by any SDK that attempts to do so.

Culprits:
- Node SDK
  - `HashiCorpVaultConfigStore` always attempts to parse upserted data as JSON in the `upsert` method. When the value is a string, this will always throw. This can be easily replicated by attempting to upsert values to the root set, since `KeyValueConfigStore` always stores keys as the secret name and the value as is when dealing with the root set.
  - The `TestCommand` will always fail since it attempts to store a number value under a secret named `CONFIGU_TEST`. 
- CLI - inherits the issue from the Node SDK 
- Python SDK - Implements `KeyValueConfigStore` similarly to the Node SDK, the issue will be the exact same there.
- Go SDK - Hashicorp vault store isn't implemented yet, not relevant here.

We need to consider how to handle key-value stores that do not support storing values as plain strings since there may be more problematic stores other than `HashiCorpVaultConfigStore`

**To Reproduce**

Attempt to use TestCommand or upsert values to root set when using `HashiCorpVaultConfigStore` in any SDK or the CLI.

**Expected behavior**

This is a tricky question that requires thought but at the very least we could store the value as an object where the key is the same key that the value came with.

**Additional context**

[Link to create/update API that states that `data` must always be a map](https://developer.hashicorp.com/vault/api-docs/v1.15.x/secret/kv/kv-v2#create-update-secret)

akash7mishra7:
Resolves: #256

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,6 +6,10 @@
     "": {
       "name": "@configu/configu",
       "license": "Apache-2.0",
+      "dependencies": {
+        "@iarna/toml": "^2.2.5",
+        "xml2js": "^0.6.2"
+      },
       "devDependencies": {
         "concurrently": "^8.2.1",
         "husky": "^8.0.3",
@@ -66,6 +70,11 @@
       "resolved": "https://registry.npmjs.org/@glideapps/ts-necessities/-/ts-necessities-2.1.3.tgz",
       "integrity": "sha512-q9U8v/n9qbkd2zDYjuX3qtlbl+OIyI9zF+zQhZjfYOE9VMDH7tfcUSJ9p0lXoY3lxmGFne09yi4iiNeQUwV7AA==",
       "dev": true
+    },
+    "node_modules/@iarna/toml": {
+      "version": "2.2.5",
+      "resolved": "https://registry.npmjs.org/@iarna/toml/-/toml-2.2.5.tgz",
+      "integrity": "sha512-trnsAYxU3xnS1gPHPyU961coFyLkh4gAD/0zQ5mymY4yOZ+CYvsPqUbOFSw0aDM4y0tV7tiFxL/1XfXPNC6IPg=="
     },
     "node_modules/@jridgewell/resolve-uri": {
       "version": "3.1.1",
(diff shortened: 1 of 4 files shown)


Sadaf-A:
Hey @akash7mishra7, please change the PR title to "Fix HashiCorpVaultConfigStore upsert method not supporting string values"
and add "Resolves: #256" in your descriton.

rannn505:
@RichardAkman Can you take a look with @RonConfigu ? what are a suggested minimal approach here with HCV?

akash7mishra7:
i have changed the title,check it !

Sadaf-A:
> i have changed the title,check it !

That's great could you please also remove the quotes from description.
Resolves: #256

akash7mishra7 (commented):
resolves #269

RichardAkman:
> @RichardAkman Can you take a look with @RonConfigu ? what is a suggested minimal solution to #256 with HCV?

Clearly, we have to store values as objects, that's a limitation inherent to how Hashicorp Vault works so that's a given. When storing "root" set values, we must store those strings as an object in some shape or form. Unfortunately, this means that the object key will have to be arbitrarily chosen by us. I think there are two intuitive options:
- The key of the object will be the same key as the config
- The key will be an arbitrary constant string like how it is currently implemented in this PR.

I'm personally inclined towards the former option rather than the latter, but I'd appreciate your input on this, @RonConfigu.

RichardAkman:
@akash7mishra7, I've observed that there's a mix of changes in this PR. It appears you've made an attempt to implement the XML config store, which is commendable. However, if that's indeed the case, I kindly request that you open a separate PR for it. 🙏🏻

RonConfigu:
> > @RichardAkman Can you take a look with @RonConfigu ? what is a suggested minimal solution to #256 with HCV?
> 
> Clearly, we have to store values as objects, that's a limitation inherent to how Hashicorp Vault works so that's a given. When storing "root" set values, we must store those strings as an object in some shape or form. Unfortunately, this means that the object key will have to be arbitrarily chosen by us. I think there are two intuitive options:
> 
> * The key of the object will be the same key as the config
> * The key will be an arbitrary constant string like how it is currently implemented in this PR.
> 
> I'm personally inclined towards the former option rather than the latter, but I'd appreciate your input on this, @RonConfigu.

I agree with @RichardAkman. Let's use `The key of the object will be the same key as the config`.
@akash7mishra7, First of all thank you for contributing 💯 . much appreciated 🥇 . Please continue with Richard's suggestion and comments.

Thanks,
Ron

RichardAkman:

Issue #268, opened 2023-10-02
Implement TOML ConfigStore in Node.js SDK


pelegpor:
### Description

- Plan a Config accommodation in the TOML format.
- Add `TOMLConfigStore` class to the Node.js SDK stores.
- Consider using the `@iarna/toml` library or similar for parsing and stringifying TOML during implementation.
- We're open to other library suggestions or, if preferred, we can implement the APIs ourselves.

### Motivation

To extend Configu's available stores collection and support the TOML format, which is a popular configuration format. This ConfigStore will be integrated into the Configu CLI as the Node.js SDK is utilized internally.

### Additional context

- [TOML - Official GitHub Repository](https://github.com/toml-lang/toml)
- [Configu Config Store Documentation](https://configu.com/docs/config-store/)
- [ConfigStore TypeScript Definition](https://github.com/configu/configu/blob/main/types/ConfigStore.ts)
- [Configu Node.js SDK Stores](https://github.com/configu/configu/tree/main/ts/packages/node/src/stores)

![image](https://github.com/configu/configu/assets/50705727/6be4950c-20c8-45b6-ab6d-a88784f05f96)

rannn505:
@cc @Sadaf-A

RichardAkman:
Hi @Dev-Arhaan, thanks for your interest in contributing! Currently, @Sadaf-A is working on this, and he has opened a PR for it. If you're still interested in helping, you might want to reach out to Sadaf and see if there are any opportunities for collaboration. Your enthusiasm is much appreciated!

Sadaf-A:
> Is anyone assigned to this task? If not I would like to work on it

@Dev-Arhaan Hey, it's great to see that you want to contribute. You can join our discord server and ping me. If you are up for it we can co-author the PR to get the job done faster.

RichardAkman:
Hi @Dev-Arhaan, Sadaf is currently unavailable to continue work on his PR. Let me know if you want to take it up and I'll assign you the issue.

Maran1947:
Hi, can i tackle this issue?

Dev-Arhaan:
Hey Richard, I Saw this very late, I'd love to work on it, let me hop on discord and try to understand the problem

PR #279, opened 2023-10-06
Implement TOML ConfigStore in Node.js SDK

Sadaf-A:
Resolves: #268


RonConfigu (changes requested):
Hi @Sadaf-A 

Please see comments on CSVConfigStore re: abstract ConfigFileStore
Same applies here.

Also, correct me if i'm wrong but TOML is not JSON and can not be parsed with JSON.parse() please refer to the issue and consider the lib @rannn505 has suggested.

Thanks for contributing 🥇 

Ron

RonConfigu:
Hi @Sadaf-A,

Please review the following impl suggestion for `TomlFileConfigStore`.

Like I said, it shouldn't extend `KeyValueConfigStore`.
The following is inspired by `IniFileConfigStore` (ts/packages/node/src/stores/IniFile.ts) please use it for insperation as well.

Please read about https://toml.io/en/v1.0.0#table

- Rename the file to `TomlFile.ts`
- Install and add as dependency [`@iarna/toml`](https://www.npmjs.com/package/@iarna/toml).
- As with `IniFileConfigStore`, please rename to `TomlFileConfigStore`
- `TomlFileConfigStore.read` should read and parse the file content as TOML and return JSON.
- `TomlFileConfigStore.get` should read the file and return configs matching with set and key. You should check each value if it is an object. if so, it means it is a nested set and you should deconstruct its keys recursivly.
- `TomlFileConfigStore.set` should read the file create new configs, override existing with new values and delete if value is empty-string. buildinig the TOML recursivly like mentioned before. Please utilze TOML native dot notation in tables instead of slash.
for example: Set `a/b/c` should translate to TOML table `[a.b.c]` and vice versa

**Please make sure your branch is up-to-date with main and run tests locally to verify your code runs and behave as expected**

Here is an example of what the TOML could look like:
```javascript
const TOML = require('@iarna/toml')
const obj = TOML.parse(`
RootSetKey = 1
[SET1]
  SET1SetKey = 2
[SET1.SET2]
  SET1SET2Key = 3
`)
// obj = { RootSetKey: 1, SET1: { SET1SetKey: 2, SET2: { SET1SET2Key: 3 } } }
const str = TOML.stringify(obj)
```

Let me know if you need anything else. You can also find me on Discord (Markeitect)

RichardAkman:

Issue #151, opened 2023-07-04
Implement `CSV` `ConfigStore` in Node.js SDK


rannn505:
**Description**

- Plan a `Config` accommodation in the csv file.
- Add `CSVConfigStore` class to the Node.js SDK stores.
- A quick search indicates that [csv](https://www.npmjs.com/package/csv) library or [xlsx](https://www.npmjs.com/package/xlsx) are the most "suitable" for use during implementation.
- We would be happy to hear more suggestions, or we can implement the APIs ourselves.

**Motivation**

Extending Configu's [available stores](https://configu.com/docs/config-store/#available-stores) collection.
This ConfigStore will be made available in the Configu CLI since the Node.js SDK is used internally.

**Additional context**

- https://en.wikipedia.org/wiki/Comma-separated_values
- https://configu.com/docs/config-store/
- https://github.com/configu/configu/blob/main/types/ConfigStore.ts
- https://github.com/configu/configu/tree/main/ts/packages/node/src/stores

![image](https://github.com/configu/configu/assets/11806915/e603c3c3-703c-46d2-91a6-3c308089a8ce)

jamesL92:
Rather than a new ConfigStore, what about generalising the existing [JsonFileConfigStore](https://github.com/configu/configu/blob/main/ts/packages/node/src/stores/JsonFile.ts#L7) to a more general `FileConfigStore` that accepts a `fileFormat` initialisation parameter?

An alternative route would be to abstract a `FileConfigStore` out of the `JsonFileConfigStore`, similar to what's done with [KeyValueConfigStore](https://github.com/configu/configu/blob/main/ts/packages/ts/src/stores/KeyValue.ts#L5)

rannn505:
@jamesL92 You've got a good point there. While the idea of a fileFormat option in the constructor sounds appealing for its generality, it may indeed lead to a bloated class due to the need to handle different formats within the same class.

With that in mind, your suggestion of an abstract FileConfigStore class is particularly attractive. This would allow us to maintain a lean, clean class while still achieving the flexibility we desire.

Let's move in this direction, and as always, thanks for your valuable insights!

cstayyab:
@rannn505 What if instead of using `FileConfigStore`  as base class we use  `JsonFileConfigStore` as base since a lot of formats can be easily represented into JSON including but not limited to `CSV`(this issue), `yaml`, `ini` (#152) and many others.

rannn505:
[...]

PR #280, opened 2023-10-06
Implement CSV ConfigStore in Node.js SDK

Sadaf-A:
Resolves: #151


RonConfigu (changes requested):
Hi @Sadaf-A,

Thank you for contributing once more 💯 

As mentioned in the comments by @rannn505 and @jamesL92 on the linked issue and the issue itself, There are a few things to change before this PR could be merged.

- Please use `csv` package like Ran asked.
- eslint rules should not be ignored and overridden with comments.

Please make sure to read the issue and its comments again before making changes.

Thanks again 🥇 

Ron

RonConfigu:
Hi @Sadaf-A,

Please review the following impl suggestion for `CsvFileConfigStore`.

Like I said, it shouldn't extend `KeyValueConfigStore`.
The following is inspired by `JsonFileConfigStore` (ts/packages/node/src/stores/JsonFile.ts) please use it for insperation as well.

- Replace `csv-parser` package with [`csv`](https://www.npmjs.com/package/csv) package. the latter is much bigger and stable.
- As with `JsonFileConfigStore`, please rename to `CsvFileConfigStore`
- `CsvFileConfigStore.read` should read and parse the file content, iterate the first 3 colunms where the first should correspond to `set`, the second is `key` and the third is `value`. These together are `Config` and this method should return an array of them.
Your impl currently returns only the key and the value without a set. This will only work if we didn't support nested sets. ie. /set1/set2
- `CsvFileConfigStore.get` should read the file and return configs matching with set and key. Your impl currently only looking for the key.
- `CsvFileConfigStore.set` should read and parse the file and its configs. append new configs, override values for existing configs and delete configs with empty-string value and write the file back to the file-system. **In this order**

**Please make sure your branch is up-to-date with main and run tests locally to verify your code runs and behave as expected**

Here is an example of what the CSV could look like. This example has headers but they will not exist on this impl:
Clipboard_2023-10-10-14-54-08

Let me know if you need anything else. You can also find me on Discord (Markeitect)

RichardAkman:

PR #295, opened 2023-10-26
Implement `FileConfigStore`

### LINKED ISSUE #156: FileStorage stores creates the file if not exists
**Description**
With the new `FileStorageConfigStore`(temp name for ref) CSV and JSON and other files should be created by the store if the file doesn't exists

**Motivation**
This will help the user with the structure of the file. for example, CSV will most likely have default headers and JSON will be created with empty array

**Additional context**
https://github.com/configu/configu/issues/151
https://github.com/configu/configu/issues/151#issuecomment-1619782423

# Tasks

- [x] TS/Node SDK implementation
- [ ] Python implementation
- [ ] Go implementation

RichardAkman:
Fixes #156 for the Node SDK, the [applied strategy is detailed in the issue](https://github.com/configu/configu/issues/156#issuecomment-1781289340). The JSON and INI config stores have been adjusted to extend the new abstract `FileConfigStore`.

@RonConfigu, we need to apply the fix to Python too. Please review the strategy and the diffs and let's implement this for Python.


RichardAkman:
The points that we discussed were applied @rannn505 @RonConfigu.

- `ExistsSync` was changed to use `fs.access`
- `read` and `write` aren't abstract anymore, these methods now expect `parseFileContent` and `stringifyConfigs` respectively to be implemented by any file store that extends the abstract class. Now, contributors are expected to only implement the specific logic that converts the file content to `Config[]` and converts `Config[]` to a string that the specific file store expects.
- A test file was added `File.test.ts` which currently only tests the init() logic. I thought about adding tests that check the `set` and `get` method but since this class is abstract it's unclear what the content of the test file should look like. I think these should be tests that are implemented per specific store implementations.
- I've checked other file stores, and a lot of them don't have packages that perform the parsing on their own. [CSV parser](https://github.com/mafintosh/csv-parser) is an exception, it does return objects but it doesn't handle writes. In this specific case, I think the CSV store should still extend this abstract class override the read method and also implement the parse method which only has a comment that says it's not used due to the circumstances.

ts/packages/node/src/stores/IniFile.test.ts
```diff
@@ -21,26 +21,20 @@ key=value
...
-    const configs = await store.read();
-    expect(configs).toEqual([
-      { set: '', key: 'rootkey', value: 'rootvalue' },
-      { set: 'section', key: 'key', value: 'value' },
-    ]);
-  });
-
   test('should write configurations to INI file', async () => {
     const store = new IniFileConfigStore({ path: testIniFilePath });
-    const newConfigs = [
```
rannn505, line 35: למנ לשנות את השמות סתם?

ts/packages/node/src/stores/IniFile.test.ts
```diff
@@ -21,26 +21,20 @@ key=value
     await fs.unlink(testIniFilePath);
   });
 
-  test('should read configurations from INI file', async () => {
```
rannn505, line 24: צריך טסט אלטרנטיבי לזה או שיהיה ברור שהטסט ההינית זה כולל גם אותו, אתה לא יכול להשאיר את הכותרת אותו הדבר!

ts/packages/node/src/stores/File.ts
```diff
@@ -0,0 +1,66 @@
...
+export abstract class FileConfigStore extends ConfigStore {
+  readonly path: string;
+  readonly initialFileState: string;
+  constructor(type: string, path: string, initialFileState: string) {
+    super(type);
+    this.path = path;
+    this.initialFileState = initialFileState;
+  }
+
+  // * Creates the file with the required "empty state" in case it does not exist
```
rannn505, line 14: remove all comments!

ts/packages/node/src/stores/File.ts
```diff
@@ -0,0 +1,66 @@
...
+  protected abstract stringifyConfigs(nextConfigs: Config[]): string;
+
+  // * Writes the next state of the configs to the file
+  private async write(nextConfigs: Config[]): Promise<void> {
+    const nextFileContent = this.stringifyConfigs(nextConfigs);
+    await fs.writeFile(this.path, nextFileContent);
+  }
+
+  async get(queries: ConfigStoreQuery[]): Promise<Config[]> {
+    const storedConfigs = await this.read();
```
rannn505, line 47: - const fileContents = read()
- const storedConfigs = parse()

ts/packages/node/src/stores/File.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  async set(configs: Config[]): Promise<void> {
+    const storedConfigs = await this.read();
+
+    const nextConfigs = _([...configs, ...storedConfigs])
+      .uniqBy((config) => `${config.set}.${config.key}`)
+      .filter((config) => Boolean(config.value))
+      .value();
+
+    await this.write(nextConfigs);
```
rannn505, line 64: - const fileContents = stringify
- write(...)

ts/packages/node/src/stores/File.ts
```diff
@@ -0,0 +1,66 @@
+import { promises as fs } from 'fs';
+import { type Config, ConfigStore, type ConfigStoreQuery } from '@configu/ts';
+import _ from 'lodash';
+
+export abstract class FileConfigStore extends ConfigStore {
+  readonly path: string;
+  readonly initialFileState: string;
+  constructor(type: string, path: string, initialFileState: string) {
```
rannn505, line 8: חליטו על יישור קו בפרמטרים לאבסטרקט קלאסס!!

ts/packages/node/src/stores/File.ts
```diff
@@ -0,0 +1,66 @@
...
+  protected abstract parseFileContent(fileContent: string): Config[];
+
+  private async read(): Promise<Config[]> {
+    const fileContent = await fs.readFile(this.path, 'utf8');
+    const parsedFileContent = this.parseFileContent(fileContent);
+    return parsedFileContent;
+  }
+
+  // * Stringifies the configs into the format the file store expects
+  protected abstract stringifyConfigs(nextConfigs: Config[]): string;
```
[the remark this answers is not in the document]
RichardAkman: done

RonConfigu (approved):
Perfect 🥇

PR #299, opened 2023-10-31
Implement eval command to upsert command piping

RichardAkman:
This PR extends the TS upsert command API to support piping of an eval command result to an upsert command.

- The Upsert command now accepts a new property called `pipe` that is a result of an eval command. If provided, the eval result is merged together with the provided configs (which take priority over an eval result). Any key that does not belong to the schema received in the upsert command is filtered, same goes for any key that is a template in the provided schema.
- The Eval and Export commands have been changed to align with the new `pipe` parameter by also having their relevant parameters renamed to `pipe` (breaking change), the affected code and docs have been adjusted accordingly.
- Also fixes a discovered unrelated bug where upserting `false` as a value does not get validated due to an erroneous condition.

Notes:
- This is a breaking change
- Requires a follow up change to the CLI
- Requires a follow up adjustment in the Python and Go SDKs


ts/packages/ts/src/commands/UpsertCommand.ts
```diff
@@ -5,65 +5,96 @@ import { ConfigError } from '../utils';
...
-        if (value) {
-          if (cfgu.template) {
-            throw new ConfigError(
-              'invalid config value',
-              `keys declared with template mustn't have a value`,
-              errorScope,
-            );
-          }
+    if (_.isEmpty(configs) && _.isEmpty(pipe)) {
+      return;
```
RonConfigu, line 61: Why would silence the Error in that case?

ts/packages/ts/src/commands/UpsertCommand.ts
```diff
@@ -5,65 +5,96 @@ import { ConfigError } from '../utils';
...
+        const schemaCfgu = schema.contents[key];
+        // * Filters keys from pipe that are not declared on the schema
+        if (!schemaCfgu) {
+          return false;
         }
+        // * Filters keys that are declared as a template in the provided schema
+        return !schemaCfgu.template;
+      })
+      .map(([key, value]) => {
+        this.validateConfigValue({ key, value: value.result.value, isFromPipe: true });
```
RonConfigu, line 78: If this is here for extra-validation, ok. If we assume pipe is from EvalResult this is redundant.

RonConfigu (commented):
Beautiful 🥇 
Please review my questions/comments

ts/packages/ts/src/commands/UpsertCommand.ts
RichardAkman: Extra validation, we want to ensure that the eval result matches the cfgu declarations of the provided schema.

ts/packages/ts/src/commands/UpsertCommand.ts
RichardAkman: Works the same like providing an empty config object, does nothing.

RonConfigu (approved):
LGTM 💯

ts/packages/ts/src/commands/EvalCommand.ts
```diff
@@ -27,7 +27,8 @@ export type EvalCommandParameters = {
   schema: ConfigSchema;
   configs?: { [key: string]: string };
   validate?: boolean;
-  previous?: EvalCommandReturn;
+  // * Allows merging the result of an eval command with the current eval result
```
rannn505, line 30: אין צורך בקומנטים האלה בבקשה תמחוק אותם מכל הפיארים שלך!

rannn505 (changes requested):
fix

ts/packages/ts/src/commands/EvalCommand.ts
RichardAkman: done

ts/packages/ts/src/commands/UpsertCommand.ts
```diff
@@ -5,65 +5,96 @@ import { ConfigError } from '../utils';
...
+      .entries()
+      .map<Config>(([key, value]) => {
+        this.validateConfigValue({ key, value, isFromPipe: false });
         return {
           set: set.path,
           key,
           value,
         };
       })
+      .unionBy(pipeConfigs, 'key')
```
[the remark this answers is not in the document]
RichardAkman: done

ts/packages/ts/src/commands/UpsertCommand.ts
```diff
@@ -66,6 +101,8 @@ export class UpsertCommand extends Command<void> {
       })
       .value();
 
-    await store.set(upsertConfigs);
+    const configsToUpsert = _.unionBy(upsertConfigs, pipeConfigs, 'key');
```
rannn505, line 104: not a good naming here! this is a the same name as in the :92.
MergedConfigsToUpsert

ts/packages/ts/src/commands/UpsertCommand.ts
```diff
@@ -5,59 +5,94 @@ import { ConfigError } from '../utils';
...
 
-        if (value) {
-          if (cfgu.template) {
-            throw new ConfigError(
-              'invalid config value',
-              `keys declared with template mustn't have a value`,
-              errorScope,
-            );
-          }
+    if (_.isEmpty(configs) && _.isEmpty(pipe)) {
```
rannn505, line 67: why not sppliting it via wach one of the loops?

ts/packages/ts/src/commands/UpsertCommand.ts
```diff
@@ -5,59 +5,94 @@ import { ConfigError } from '../utils';
...
+          key,
+          value: value.result.value,
+        };
+      })
+      .value();
 
+    const upsertConfigs = _(configs)
+      .entries()
+      .map<Config>(([key, value]) => {
+        this.validateConfigValue({ key, value, skipDeclarationAndTemplateValidation: false });
```
rannn505, line 95: I would convert the whole mapper function here

rannn505: split the skip flag to each of the validation

rannn505 (changes requested):
please see comments and fix

ts/packages/ts/src/commands/UpsertCommand.ts
RichardAkman: I don't understand the question. This condition is to skip the async init method in case there is nothing to upsert.

ts/packages/ts/src/commands/UpsertCommand.ts
RichardAkman: I split the validations. I don't understand what "convert the whole mapper function here" means.. please elaborate.

rannn505 (approved):

PR #300, opened 2023-10-31
Implement OpenFeatureConfigStore and related stores

RonConfigu:
- [x] LaunchDarklyConfigStore
- [x] CloudBeesConfigStore


ts/packages/node/src/index.ts
```diff
@@ -11,6 +11,8 @@ export * from './stores/HashiCorpVault';
 export * from './stores/Keyv';
 export * from './stores/KubernetesSecret';
 export * from './stores/LaunchDarkly';
```
rannn505, line 13: No need for the open feature prefix just set those the main launchdarklyconfigstore same for all others. Same for filenames and types. Remove the old one.

ts/packages/node/src/stores/OpenFeature.ts
```diff
@@ -0,0 +1,69 @@
...
+        if (details.errorCode === 'TYPE_MISMATCH') return this.getValue(key, set, '');
+        return JSON.stringify(details.value);
+      default:
+        return '';
+    }
+  }
+
+  async get(queries: ConfigStoreQuery[]): Promise<Config[]> {
+    const queryPromises = queries.map(async ({ set, key }) => {
+      const value = await this.getValue(key, set);
```
rannn505, line 59: Where are you iterating each toe? You always using the book default

ts/packages/node/src/stores/OpenFeature.ts
```diff
@@ -0,0 +1,69 @@
...
+export abstract class OpenFeatureConfigStore extends ConfigStore {
+  protected readonly provider: Provider | Promise<Provider>;
+  private readonly context: EvaluationContext;
+  private readonly client: Client;
+
+  protected constructor(type: string, configuration: OpenFeatureConfigStoreConfiguration) {
+    super(type);
+    const { provider, context = {} } = configuration;
+    this.provider = provider;
+    this.context = context;
```
rannn505, line 24: No need for saving the context. Set it globally on the client and let them handle the overriding and merging.

rannn505 (changes requested):
See comments

ts/packages/node/src/index.ts
RonConfigu: Sure. this means breaking changes for any `LaunchDarklyConfigStore` since store.set() is disabled.
we're ok with that?

ts/packages/node/src/stores/OpenFeature.ts
RonConfigu: sorry, didn't understand the question

ts/packages/node/src/stores/OpenFeature.ts
RonConfigu: np

ts/packages/node/src/index.ts
```diff
@@ -6,6 +6,7 @@ export * from './stores/IniFile';
 export * from './stores/AWSParameterStore';
 export * from './stores/AWSSecretsManager';
 export * from './stores/AzureKeyVault';
+export * from './stores/CloudBees';
```
rannn505, line 9: למה לדחוף את זה באמצע

RonConfigu: Thought the list should be A-Z..
I'll move it to the end

ts/packages/node/src/stores/OpenFeature.ts
```diff
@@ -0,0 +1,69 @@
...
+    const queryPromises = queries.map(async ({ set, key }) => {
+      const value = await this.getValue(key, set);
+      return { set, key, value };
+    });
+    const configs = await Promise.all(queryPromises);
+    return configs.filter((config) => config.value !== '');
+  }
+
+  set(configs: Config[]): Promise<void> {
+    throw new Error(`OpenFeatureConfigStore doesn't support upsert`);
```
RonConfigu, line 65: @rannn505 @RichardAkman 
Let me know if you want different error message. Also, should I use `ConfigError` in this case? in any case?
Thanks

RichardAkman (approved):
LGTM 🥇

rannn505:
why the branch starting with 0- and not the real number of the issue?

ts/packages/node/package.json
```diff
@@ -42,7 +42,11 @@
     "@configu/ts": "*",
     "@google-cloud/secret-manager": "^5.0.1",
     "@kubernetes/client-node": "^0.19.0",
+    "@launchdarkly/node-server-sdk": "^9.0.1",
```
rannn505, line 45: this is not ordered!

ts/packages/node/src/stores/CloudBees.ts
```diff
@@ -0,0 +1,23 @@
...
+  context?: EvaluationContext;
+};
+
+export class CloudBeesConfigStore extends OpenFeatureConfigStore {
+  constructor(configuration: CloudBeesConfigStoreConfiguration) {
+    const { appKey, providerOptions, context } = configuration;
+    super('cloud-bees', {
+      provider: CloudbeesProvider.build(appKey, {
+        ...providerOptions,
+        fetchIntervalInSec: 0,
```
rannn505, line 18: why?

rannn505: add comment to explain why user cant override it

ts/packages/node/src/stores/LaunchDarkly.ts
```diff
@@ -1,174 +1,34 @@
...
+
+export class LaunchDarklyConfigStore extends OpenFeatureConfigStore {
+  constructor(configurations: LaunchDarklyConfigStoreConfiguration) {
+    const { sdkKey, ldOptions, context } = configurations;
+    if (_.isEmpty(context) || !['targetingKey', 'key'].some((value) => Object.keys(context).includes(value)))
+      throw new Error(
+        `The EvaluationContext must contain either a 'targetingKey' or a 'key' and the type must be a string.`,
+      );
+    // * stream is set to true by default
+    const ldOptionsWithoutStream = { ...ldOptions, stream: false };
```
rannn505, line 22: add comment to explain why user cant override it

rannn505 (approved):

ts/packages/node/src/stores/OpenFeature.test.ts
```diff
@@ -0,0 +1,97 @@
+import { InMemoryProvider, type EvaluationContext } from '@openfeature/server-sdk';
+import { ConfigSet, ConfigSchema, EvalCommand, ExportCommand, type EvalCommandReturn } from '@configu/ts';
```
rannn505, line 2: why testing a store require Commands? this is not e2e tests those are unit tests, you hsould use only the class itself with the get and the set method. please insert a fix
@RonConfigu @RichardAkman

PR #324, opened 2023-12-11
Configu cli next

### LINKED ISSUE #247: EvalCommand prioritizes current schema defaults over previous EvalCommand result
**Actual behavior**

When using eval command chaining, the previous result is always overridden by the current default value from the schema. This severely harms the usability of eval command chaining.

**To Reproduce**

- Create a schema that has a key with a default value. (the `get-started` schema works for this example)
- Upsert a value to a key with a default value to any store (use `GREETING` from the `get-started` schema)
- Use eval command that gets the previously upserted value for the key with a default value and then forward the result to another eval command that uses the default value in the result.
- View the result of the final eval command and see that the result for the key is always the default value and not the value that was fetched from stores.

**Expected behavior**

Config values from stores should always be prioritized over default values.

**Screenshots**

image

If the default of the key is removed, the expected result occurs:
image

### LINKED ISSUE #297: Implement eval command to upsert command piping
**Description**

Upsert command will accept the result of Eval command as a parameter and will be used to perform the upsert.

**Motivation**

This will improve the usability of both Eval and Upsert commands in general. A very nice use case is migration between different stores.

## Tasks

- [x] TS/Node SDK implementation
- [x] CLI implementation

rannn505:
- upgrade oclif to v3
- upgrade all cli deps
- upgrade configu sdk - use the new schema syntax with a wrapper getSchemaInstanceBySchemaFlag
- allow yaml schema (experimental)
- upgrade cosmiconfig major v9 - searchStrategy
- extract common functions as utils
- add `launch-darkly`, `cloud-bees` to cli store dict
- add pipe to upsert
- fix upsert --import fails on extra/missing keys from schema
- fix run command cwd - allow customize, default to .configu location
- init - allow override name always even on premade schemas
- init - new initial state "some key/value"

Closes #247 
Closes #297


RichardAkman (commented):
Looks great, the upsert command piping works well and I also played around with the YAML schema format which looks solid. I haven't looked too deep into the new OClif major version, I assume you verified that there are no breaking changes.

Issue #343, opened 2024-01-18
Adjust `ConfiguConfigStore` for the `Approval Queue` feature


RichardAkman:
### Suggestion

The `ConfiguConfigStore` class `set` method currently doesn't acknowledge that configs might not have been mutated but instead inserted into an approval queue for admin review. To indicate to users of the store, the method should throw an exception in this scenario.

### Motivation

Improved UX and clarity to users of `ConfiguConfigStore`

### Tasks

- [x] TS SDK implementation
- [x] CLI implementation

RichardAkman:
This issue is no longer relevant due to the Python and Go SDKs being deprecated in favor of `@configu/proxy`

PR #344, opened 2024-01-18
Adjust `@configu/ts` - `ConfiguConfigStore` for the Approval Queue feature

RichardAkman:
Part of #343


ts/packages/cli/package.json
```diff
@@ -98,8 +98,8 @@
   ],
   "dependencies": {
     "@configu/lib": "^0.7.0",
-    "@configu/node": "^0.20.1",
-    "@configu/ts": "^0.20.1",
+    "@configu/node": "^0.20.2-next.4",
+    "@configu/ts": "^0.20.2-next.4",
```
RichardAkman, line 102: TODO: remove this before merging

ts/packages/ts/src/stores/Configu.ts
```diff
@@ -36,6 +48,14 @@ export class ConfiguConfigStore extends ConfigStore {
   }
 
   async set(configs: Config[]): Promise<void> {
-    await this.client.put('/config', { configs });
+    const response = await this.client.put('/config', { configs });
+    if (response.status === 202) {
+      const protectedSet = response.data.diff.pending[0].set;
+      const queueUrl = `${response.data.queueUrl}?sp=${protectedSet}`;
+      throw new ConfiguConfigStoreApprovalQueueError(
```
rannn505, line 55: This error building should be made in the very specific error type you created `ConfiguConfigStoreApprovalQueueError`.
Personally I prefer just a "normal error here" as I can seem to think how should it be reused anywhere else.

ts/packages/ts/src/stores/Configu.ts
```diff
@@ -9,8 +9,20 @@ export type ConfiguConfigStoreConfiguration = {
...
+
+  constructor(message: string, queueUrl: string) {
+    super(message);
+    this.queueUrl = queueUrl;
+  }
+}
+
 export class ConfiguConfigStore extends ConfigStore {
   private client: Axios;
+  public static readonly approvalQueueErrorName = ConfiguConfigStoreApprovalQueueErrorName;
```
rannn505, line 25: dont see any point of that

rannn505 (changes requested):
@RichardAkman see comments please

ts/packages/ts/src/stores/Configu.ts
RichardAkman: - Good point regarding the building of the error
- We need a unique error otherwise we will have to rely on the error message string to identify if this error was triggered by the approval queue and we don't want to do that.

ts/packages/ts/src/stores/Configu.ts
RichardAkman: using `error instanceof ConfiguConfigStoreApprovalQueueError` is unreliable, it gives us `false` when used through the browser SDK in the SaaS. So instead we check for the error via the `name` property which we export outwards through this class property. If we export the error itself we'd need to change the browser/node SDK to also export the errors and we didn't wanna change those SDKs. So this is the best solution we found so far, if you have any better ideas or input, we'd love it hear it.

ts/packages/cli/src/commands/upsert.ts
RichardAkman, line 1: TODO: split into another follow-up PR

ts/packages/ts/src/stores/Configu.ts
```diff
@@ -9,8 +9,22 @@ export type ConfiguConfigStoreConfiguration = {
   source?: string;
 };
 
+const ConfiguConfigStoreApprovalQueueErrorName = 'ConfiguConfigStoreApprovalQueueError';
```
rannn505, line 12: @RichardAkman sorry but I still cant accept this change, its too wired, your only option is to use instanceOf if your going with the Custom Error class.

rannn505 (changes requested):
@RichardAkman see comments

rannn505 (approved):

PR #345, opened 2024-01-18
Implement ExportCommand Key Mutation Callback

### LINKED ISSUE #333: ExportCommand Key Mutation Callback
### Suggestion

Introduce a new optional property of the `ExportCommand` called `keys` where users can provide a callback function `(key: string) => string`. This function will take a `configKey` as a string, apply desired mutations (like adding a prefix or suffix), and return a modified key to be used in the export result.

### Motivation

- This addition will enable functionalities such as `--prefix` or `--suffix` in the export command of the CLI, allowing for more versatile command options.
- Adapting keys on export is a common practice in application configuration management, enhancing flexibility for different environments.
- It accommodates specific requirements of some compilers or builders that necessitate particular prefixes for environment variables.

### Context

_No response_

RonConfigu:
## QA Report I
- [x] Test Units Passes

Closes #333


ts/packages/ts/src/commands/ExportCommand.ts
```diff
@@ -1,22 +1,44 @@
...
+
+    if (!keys) {
+      return result;
+    }
+
+    return _.mapKeys(result, (current, key) => {
+      try {
+        const mutatedKey = keys(key);
+        if (!NAME(mutatedKey)) {
+          throw new ConfigError('invalid config key', `${mutatedKey} must be a valid name`);
```
RichardAkman, line 31: Error message alignment with `ConfigSchema` key validation.

```suggestion
          throw new ConfigError('invalid config key', `key "${mutatedKey}" mustn't contain reserved words`);
```

ts/packages/ts/src/commands/ExportCommand.ts
```diff
@@ -1,22 +1,44 @@
...
+        }
+        return mutatedKey;
+      } catch (error) {
+        return key;
```
RichardAkman, line 35: Why are we returning the original key and not throwing an exception? There is no requirement for that in the issue and in terms of UX, it makes things very unclear for the user if they supply a function that doesn't pass `NAME` validation.

ts/packages/ts/src/commands/ExportCommand.ts
```diff
@@ -1,22 +1,44 @@
...
+      return result;
+    }
+
+    return _.mapKeys(result, (current, key) => {
+      try {
+        const mutatedKey = keys(key);
+        if (!NAME(mutatedKey)) {
+          throw new ConfigError('invalid config key', `${mutatedKey} must be a valid name`);
+        }
+        return mutatedKey;
```
RichardAkman, line 33: Let's maintain the error scoping in usage of the commands. This also aligns with the same logic in `ConfigSchema` key validation:
```node.js
  const errorScope: [string, string][] = [
        ['ExportCommand', ''],
        ['ConfigKey', key],
      ];
      const mutatedKey = keys(key);
      if (!NAME(mutatedKey)) {
        throw new ConfigError('invalid config key', `${mutatedKey} must be a valid name`, errorScope);
      }
      return mutatedKey;
```

ts/packages/ts/src/commands/commands.test.ts
RichardAkman, line 1: I'm kind of conflicted regarding any test that breaks the signature of `keys`. These tests cover scenarios where the user wants to break our interface intentionally as Typescript already validates this for the user of the SDK. I'm leaning towards them being redundant and unnecessary for the same reason we don't check the `typeof` of the return of `keys` being a string.

ts/packages/ts/src/commands/commands.test.ts
```diff
@@ -222,4 +223,162 @@ describe(`commands`, () => {
...
+          },
+        }),
+        configs: {
+          KEY0: 'KEY0',
+          KEY1: 'KEY1',
+        },
+      }).run();
+
+      const exportedConfigs = await new ExportCommand({ pipe: evalResult, keys: (key) => `!${key}` }).run();
+      expect(exportedConfigs).toStrictEqual({ KEY0: 'KEY0', KEY1: 'KEY1' });
```
RichardAkman, line 355: ExportCommand should throw in this scenario, this test should be adjusted

ts/packages/ts/src/commands/commands.test.ts
```diff
@@ -222,4 +223,162 @@ describe(`commands`, () => {
...
+        },
+      }).run();
+
+      const exportedConfigs = await new ExportCommand({
+        pipe: evalResult,
+        keys: (key) => {
+          throw new Error('bad key');
+        },
+      }).run();
+      expect(exportedConfigs).toStrictEqual({ KEY0: 'KEY0', KEY1: 'KEY1' });
```
RichardAkman, line 381: Should expect an exception

ts/packages/ts/src/commands/ExportCommand.ts
RonConfigu: There is no requirement to raise exception as well.

Refactored

RonConfigu: Test refactored accordingly

ts/packages/ts/src/commands/ExportCommand.ts
RonConfigu: done

ts/packages/ts/src/commands/commands.test.ts
RonConfigu: done

ts/packages/ts/src/commands/commands.test.ts
```diff
@@ -222,4 +223,162 @@ describe(`commands`, () => {
...
+          KEY1: 'KEY1',
+        },
+      }).run();
+
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-expect-error
+      const exportedConfigs = await new ExportCommand({ pipe: evalResult, keys: (key) => 5 }).run();
+      expect(exportedConfigs).toStrictEqual({ KEY0: 'KEY0', KEY1: 'KEY1' });
+    });
+    test('Export with bad keys mutation callback - returns empty string', async () => {
```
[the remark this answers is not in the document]
RonConfigu: so? what if NAME will break or change?

ts/packages/ts/src/commands/commands.test.ts
RonConfigu: done

ts/packages/ts/src/commands/commands.test.ts
RonConfigu: better safe than sorry in my opinion.
If you want i'll delete these tests

RonConfigu:
@RichardAkman Please review the code again. changes have been made to the command not only the tests.

ts/packages/ts/src/commands/commands.test.ts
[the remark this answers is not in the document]
RichardAkman: Fair enough, extra tests don't hurt.

ts/packages/ts/src/commands/commands.test.ts
RichardAkman: Fair enough, extra tests don't hurt.

ts/packages/ts/src/commands/commands.test.ts
```diff
@@ -222,4 +224,179 @@ describe(`commands`, () => {
...
+        configs: {
+          KEY0: 'KEY0',
+          KEY1: 'KEY1',
+        },
+      }).run();
+      await expect(
+        new ExportCommand({
+          pipe: evalResult,
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-expect-error
```
RichardAkman, line 292: NIT: Cleans up the disable next line. Check if it works for you.
```suggestion
          // @ts-expect-error - should throw ConfigError
```

ts/packages/ts/src/commands/commands.test.ts
```diff
@@ -222,4 +224,179 @@ describe(`commands`, () => {
...
```
RichardAkman, line 319: NIT: Cleans up the disable next line. Check if it works for you.
```suggestion
          // @ts-expect-error - should throw ConfigError
```

RonConfigu:
@RonConfigu Fix test to use method for evals

ts/packages/ts/src/commands/ExportCommand.ts
```diff
@@ -1,22 +1,48 @@
...
+      } catch (error) {
+        throw new ConfigError('invalid config key', error.message, errorScope);
+      }
+    });
+  }
+
   async run() {
     const { pipe } = this.parameters;
-    return _.mapValues(pipe, (current) => current.result.value);
+    return _.mapValues(this.mutateKeys(pipe), (current) => current.result.value);
```
rannn505, line 46: For better debugging experience

```suggestion
    const configDict = _.mapValues(pipe, (current) => current.result.value);
    const keyMutatedConfigDict = this.mutateKeys(configDict);
    return keyMutatedConfigDict;
```

ts/packages/ts/src/commands/commands.test.ts
```diff
@@ -222,4 +224,172 @@ describe(`commands`, () => {
       }
     });
   });
+  describe(`ExportCommand`, () => {
```
rannn505, line 227: Looks great!

ts/packages/ts/src/commands/commands.test.ts
```diff
@@ -222,4 +224,172 @@ describe(`commands`, () => {
...
+          KEY0: 'KEY0',
+          KEY1: 'KEY1',
+        },
+      }).run();
+      await expect(
+        new ExportCommand({
+          pipe: evalResult,
+          keys: (key) => '',
+        }).run(),
+      ).rejects.toBeInstanceOf(ConfigError);
```
rannn505, line 342: not sure is it returning an empty string or throwing an error?
`''` is not in the NAME `RESERVED_NAMES` list

rannn505 (changes requested):
@RonConfigu thats gret, take a look at my comments 🙏🏽

ts/packages/ts/src/commands/ExportCommand.ts
```diff
@@ -1,22 +1,57 @@
...
+        ['ExportCommand', ''],
+        ['ConfigKey', key],
+      ];
+      let mutatedKey = '';
+      let isKeyValid = true;
+      try {
+        mutatedKey = String(keys(key));
+      } catch (error) {
+        isKeyValid = false;
+      }
```
RichardAkman, line 38: I'm not sure why we are masking errors that come from `keys`, unless there's a reason I'm not aware of, I think we should let the user see the error that comes from their own function

ts/packages/ts/src/commands/commands.test.ts
```diff
@@ -222,4 +224,81 @@ describe(`commands`, () => {
...
+    });
+    test('Export with bad keys mutation callback - returns number', async () => {
+      const evalResult = await getEvalResult();
+      // @ts-expect-error - should throw ConfigError
+      const exportedConfigs = await new ExportCommand({ pipe: evalResult, keys: (key) => 5 }).run();
+      expect(exportedConfigs).toStrictEqual({ '5': 'KEY1' });
+    });
+    test('Export with bad keys mutation callback - returns empty string', async () => {
```
RichardAkman, line 274: `returns empty string` isn't true, it throws, please adjust the test name 🙏🏻

ts/packages/ts/src/commands/ExportCommand.ts
RonConfigu: done

ts/packages/ts/src/commands/commands.test.ts
RonConfigu: done

RichardAkman (approved):
LGTM 💯

rannn505 (approved):
@RonConfigu thanks, I wil go on to work on the CLI issue now

PR #347, opened 2024-01-21
Implement lazy Property for Cfgu Declaration

### LINKED ISSUE #335: Implement `lazy` Property for Cfgu Declaration
### Suggestion

Adding a `lazy` boolean property to the Cfgu declaration. This property allows users to specify that a value for a given `ConfigKey` should only be set during the `eval` command execution, either using the `configs` parameter or the `--config` flag in the CLI. Essentially, `lazy: true` implies the following:
- Values for this key cannot be assigned using the `upsert` command.
- Any existing value (`Config`) for this key in any store will be disregarded during `eval`.
- The only method to assign a value to this key is through the `eval` command with the `configs` parameter.
- If marked as `required`, the user must provide a value for this key during `eval`.
- This property cannot be combined with `template` and `default` properties.

```json
# my_schema.cfgu.json
{
  ...
  "MY_EMAIL": {
    "type": "Email",
    "lazy": true
  }
  ...
} 
```

```bash
configu upsert ... --schema './my_schema.cfgu.json' -c "MY_EMAIL=ran@configu.com" # Error

configu eval --schema './my_schema.cfgu.json' -c "MY_EMAIL=ran@configu.com" # OK
```

### Motivation

This feature enhances the Cfgu declaration's versatility, catering to scenarios where configuration values need to be dynamic or context-specific. It offers greater control over when and how configuration values are assigned, particularly beneficial in environments with varying runtime contexts. By enabling lazy evaluation of config values, this feature facilitates more adaptable and responsive configuration setups, particularly useful in complex, multi-stage deployment processes or when dealing with sensitive information that requires contextual assignment.

### Context

_No response_

RonConfigu:
## QA Report I
- [x] Test Units Passes

Closes #335


types/Cfgu.ts
```diff
@@ -1,40 +1,40 @@
 export type CfguType =
```
RichardAkman, line 1: I'm guessing your IDE did this? Let's not change this 🙏🏻

ts/packages/ts/src/commands/UpsertCommand.ts
```diff
@@ -60,6 +60,10 @@ export class UpsertCommand extends Command<void> {
           );
         }
 
+        if (cfgu.lazy) {
+          throw new ConfigError('invalid config key', `key "${key}" is lazy and cannot be set directly`, errorScope);
```
RichardAkman, line 64: How about this to keep it similar to the template error:
```suggestion
          throw new ConfigError('invalid config key', `keys declared as lazy mustn't have a value`, errorScope);
```

ts/packages/ts/src/commands/EvalCommand.ts
```diff
@@ -36,7 +36,14 @@ export class EvalCommand extends Command<EvalCommandReturn> {
   }
 
   private evalFromConfigsOverride(result: EvalCommandReturn): EvalCommandReturn {
```
RichardAkman, line 38: I strongly feel like this method should remain untouched and `validateResult` should be responsible for throwing the `required` error which it already does. Is there any specific reason why we need to duplicate the validation here?

ts/packages/ts/src/commands/commands.test.ts
```diff
@@ -221,5 +243,147 @@ describe(`commands`, () => {
         await Promise.all(deletePromises);
       }
     });
+    describe('Tests for lazy configs', () => {
+      test('run EvalCommand without configs but  `cfgu.lazy && cfgu.required = true`', async () => {
```
RichardAkman, line 247:
```suggestion
      test('run EvalCommand without configs but `cfgu.lazy && cfgu.required = true`', async () => {
```

ts/packages/ts/src/commands/commands.test.ts
```diff
@@ -221,5 +243,147 @@ describe(`commands`, () => {
...
+        expect.assertions(1);
```
RichardAkman, line 248: Not sure why we need these, please explain 🙏🏻

ts/packages/ts/src/commands/commands.test.ts
```diff
@@ -221,5 +243,147 @@ describe(`commands`, () => {
...
+        await expect(() =>
+          new EvalCommand({
+            store: store1,
+            set: set1,
+            schema: new ConfigSchema('lazy', {
+              K1: {
+                type: 'String',
+                lazy: true,
+                required: true,
```
RichardAkman, line 295: the test description says `cfgu.required = false`

ts/packages/ts/src/commands/commands.test.ts
```diff
@@ -221,5 +243,147 @@ describe(`commands`, () => {
...
+              required: true,
+            },
+          }),
+          configs: {
+            K1: '1',
+          },
+        }).run();
+        expect(result).toMatchObject({ K1: { result: { value: '1' } } });
+      });
+      test('run EvalCommand without override for lazy config when `cfgu.required = false`', async () => {
```
RichardAkman, line 341: the `required` prop is not defined, we need to clarify the test names and have them align with what's going on inside the test.

ts/packages/ts/src/commands/commands.test.ts
```diff
@@ -221,5 +243,147 @@ describe(`commands`, () => {
...
+              type: 'String',
+              lazy: true,
+            },
+          }),
+        }).run();
+        expect(result).toMatchObject({ K1: { result: { value: '' } } });
+      });
+
+      test('run EvalCommand with override for lazy config when there is a value in the store and get the override value in the result', async () => {
+        await store1.set([{ set: set1.path, key: 'K1', value: '2' }]);
```
RichardAkman, line 356: I think we need to avoid any direct use of the store methods from outside the commands themselves, we don't want people to get the idea that we want them to do these things, even for tests. Prefer using upsert.

RichardAkman (changes requested):
Great work 💯 please review my comments 🙏🏻

types/Cfgu.ts
RonConfigu: Sure. sorry

ts/packages/ts/src/commands/commands.test.ts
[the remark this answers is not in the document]
RonConfigu: There are no other PR's. Please elaborate on the problem here.
Thanks

ts/packages/ts/src/commands/commands.test.ts
RonConfigu: Strongly disagree.
This test has nothing to do we UpsertCommand. If it would be possible I would imitate a store with these in it.
Generally speaking, in uint-tests like in real life, one should eliminate all variables from the experiment and only change the tested one. proving two variables in one test is not valid for either. (unless they are entangled)

ts/packages/ts/src/commands/commands.test.ts
RonConfigu: The "error" here is in the title not the prop.
the test right below has this prop = false.

Thanks

ts/packages/ts/src/commands/commands.test.ts
RonConfigu: all tests names have been refactored. please review.

ts/packages/ts/src/commands/commands.test.ts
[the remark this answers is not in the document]
RichardAkman: This code is duplicated 6 times:
```node.js
const evalResult = await new EvalCommand({
        store: store1,
        set: set1,
        schema: new ConfigSchema('mutate', {
          KEY0: {
            type: 'String',
          },
          KEY1: {
            type: 'String',
          },
        }),
        configs: {
          KEY0: 'KEY0',
          KEY1: 'KEY1',
        },
      }).run();
```

let's keep things DRY and recycle this code.

ts/packages/ts/src/commands/commands.test.ts
RichardAkman: Then at the very least remove the setup part from the test itself by using something like [`beforeAll`](https://jestjs.io/docs/setup-teardown#one-time-setup)

ts/packages/ts/src/commands/EvalCommand.ts
RonConfigu: done

ts/packages/ts/src/commands/EvalCommand.ts
```diff
@@ -291,7 +291,10 @@ export class EvalCommand extends Command<EvalCommandReturn> {
       ...(await this.evalFromStoreSet(
         _.pickBy(
           result,
-          (current) => current.result.origin === EvaluatedConfigOrigin.EmptyValue && !current.context.cfgu.template,
+          (current) =>
+            current.result.origin === EvaluatedConfigOrigin.EmptyValue &&
+            !current.context.cfgu.template &&
+            !current.context.cfgu.lazy,
```
rannn505, line 297: this should already been done on the `evalFromConfigsOverride` no?
you should validate somewhere that all lazy configs are set via the config override.

rannn505 (changes requested):
@RonConfigu great work

ts/packages/ts/src/commands/EvalCommand.ts
```diff
@@ -158,6 +171,7 @@ export class EvalCommand extends Command<EvalCommandReturn> {
 
     const mergedResults = _([pipe, result])
       .flatMap((current) => _.values(current))
+      .value()
```
RichardAkman, line 174: I remember that you had a linting issue here but it's an unnecessary change that adds more complexity to the operation and is irrelevant to this PR. Let's omit this for now 🙏🏻

ts/packages/ts/src/commands/EvalCommand.ts
```diff
@@ -37,7 +37,20 @@ export class EvalCommand extends Command<EvalCommandReturn> {
 
   private evalFromConfigsOverride(result: EvalCommandReturn): EvalCommandReturn {
     if (!this.parameters.configs) {
-      return result;
+      return _.mapValues(result, (current) => {
```
RichardAkman, line 40: The condition here is incorrect, if you have two lazy keys and one is provided and the second isn't you'll end up evaluating from the store for the missing key.

How about this direction? We let the existing override logic always run and check if there is a lazy property on the keys
```node.js
  private evalFromConfigsOverride(result: EvalCommandReturn): EvalCommandReturn {
    return _.mapValues(result, (current) => {
      const { context } = current;
      const hasConfigOverrideValue = Object.prototype.hasOwnProperty.call(this.parameters.configs, context.key);

      if (!hasConfigOverrideValue && !context.cfgu.lazy) {
        return current;
      }

      const configOverrideValue = this.parameters.configs?.[context.key] ?? '';
      return {
        ...current,
        result: {
          origin: EvaluatedConfigOrigin.ConfigsOverride,
          source: `parameters.configs.${context.key}=${configOverrideValue}`,
          value: configOverrideValue,
        },
      };
    });
  }
```

ts/packages/ts/src/commands/EvalCommand.ts
RonConfigu: done

ts/packages/ts/src/commands/EvalCommand.ts
RonConfigu: done

ts/packages/ts/src/commands/commands.test.ts
RonConfigu: done

RichardAkman (approved):
LGTM 🎖️

ts/packages/ts/src/commands/EvalCommand.ts
RonConfigu: done

ts/packages/ts/src/commands/EvalCommand.ts
rannn505: @RonConfigu image is broken here

rannn505 (approved):
@RonConfigu great work

PR #351, opened 2024-01-31
Update @configu/cli dependencies for @configu/ts, @configu/node

### LINKED ISSUE #335: Implement `lazy` Property for Cfgu Declaration

RonConfigu:
Closes #335

### DIFF
--- a/ts/package-lock.json
+++ b/ts/package-lock.json
@@ -19195,8 +19195,8 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@configu/lib": "^0.7.0",
-        "@configu/node": "^0.20.1",
-        "@configu/ts": "^0.20.1",
+        "@configu/node": "^0.21.0",
+        "@configu/ts": "^0.21.0",
         "@oclif/core": "^3.14.1",
         "@oclif/plugin-autocomplete": "^3.0.3",
         "@oclif/plugin-not-found": "^3.0.5",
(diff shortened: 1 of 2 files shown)


RichardAkman (approved):
LGTM

PR #356, opened 2024-02-08
implement xml configstore in nodejs sdk

### LINKED ISSUE #269: Implement XML ConfigStore in Node.js SDK
### Description

- Plan a Config accommodation in the XML format.
- Add `XMLConfigStore` class to the Node.js SDK stores.
- Consider using the `xml2js` library or similar for parsing and converting XML during implementation.
- We're open to other library suggestions or, if preferred, we can implement the APIs ourselves.

### Motivation

To extend Configu's available stores collection and support the XML format, which is a widely used configuration format. This ConfigStore will be integrated into the Configu CLI as the Node.js SDK is utilized internally.

### Additional context

- [XML on Wikipedia](https://en.wikipedia.org/wiki/XML)
- [Configu Config Store Documentation](https://configu.com/docs/config-store/)
- [ConfigStore TypeScript Definition](https://github.com/configu/configu/blob/main/types/ConfigStore.ts)
- [Configu Node.js SDK Stores](https://github.com/configu/configu/tree/main/ts/packages/node/src/stores)

image

Maran1947:
Closes #269 
Hi @RichardAkman, Can you please review it?

### DIFF
--- a/ts/package-lock.json
+++ b/ts/package-lock.json
@@ -4819,6 +4819,15 @@
         "@types/node": "*"
       }
     },
+    "node_modules/@types/xml2js": {
+      "version": "0.4.14",
+      "resolved": "https://registry.npmjs.org/@types/xml2js/-/xml2js-0.4.14.tgz",
+      "integrity": "sha512-4YnrRemBShWRO2QjvUin8ESA41rH+9nQGLUGZV/1IDhi3SL9OhdpNC/MrulTWuptXKwhx/aDxE7toV0f/ypIXQ==",
+      "dev": true,
+      "dependencies": {
+        "@types/node": "*"
+      }
+    },
     "node_modules/@types/yargs": {
       "version": "17.0.13",
       "dev": true,
@@ -14861,7 +14870,6 @@
     },
     "node_modules/sax": {
       "version": "1.2.1",
-      "dev": true,
       "license": "ISC"
     },
     "node_modules/scoped-regex": {
@@ -18095,7 +18103,6 @@
       "version": "11.0.1",
       "resolved": "https://registry.npmjs.org/xmlbuilder/-/xmlbuilder-11.0.1.tgz",
       "integrity": "sha512-fDlsI/kFEx7gLvbecc0/ohLG50fugQp8ryHzMTuW9vSa1GJ0XYWKnhsUx7oie3G98+r56aTQIUB4kht42R3JvA==",
-      "dev": true,
       "engines": {
         "node": ">=4.0"
       }
@@ -19781,12 +19788,14 @@
         "pg": "^8.11.3",
         "reflect-metadata": "^0.1.13",
         "sqlite3": "^5.1.6",
-        "typeorm": "^0.3.17"
+        "typeorm": "^0.3.17",
+        "xml2js": "^0.6.2"
       },
       "devDependencies": {
         "@types/ini": "^1.3.34",
         "@types/lodash": "^4.14.202",
-        "@types/rox-node": "^5.0.3"
+        "@types/rox-node": "^5.0.3",
+        "@types/xml2js": "^0.4.14"
       },
       "engines": {
         "node": ">=v16.0.0"
@@ -20004,6 +20013,18 @@
       },
       "engines": {
         "node": ">=14.0.0"
+      }
+    },
+    "packages/node/node_modules/xml2js": {
+      "version": "0.6.2",
+      "resolved": "https://registry.npmjs.org/xml2js/-/xml2js-0.6.2.tgz",
+      "integrity": "sha512-T4rieHaC1EXcES0Kxxj4JWgaUQHDk+qwHcYOCFHfiwKz7tOVPLq7Hjq9dM1WCMhylqMEfP7hMcOIChvotiZegA==",
+      "dependencies": {
+        "sax": ">=0.6.0",
+        "xmlbuilder": "~11.0.0"
+      },
+      "engines": {
+        "node": ">=4.0.0"
       }
     },
     "packages/ts": {
@@ -21829,6 +21850,7 @@
         "@types/ini": "^1.3.34",
         "@types/lodash": "^4.14.202",
         "@types/rox-node": "^5.0.3",
+        "@types/xml2js": "^0.4.14",
         "axios": "^1.6.2",
         "cloudbees-openfeature-provider-node": "^1.0.0",
         "etcd3": "^1.1.2",
@@ -21840,7 +21862,8 @@
         "pg": "^8.11.3",
         "reflect-metadata": "^0.1.13",
         "sqlite3": "^5.1.6",
-        "typeorm": "^0.3.17"
+        "typeorm": "^0.3.17",
+        "xml2js": "^0.6.2"
       },
       "dependencies": {
         "@azure/msal-browser": {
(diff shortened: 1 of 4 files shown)


Maran1947:
Sure thing, I'll go and make these changes.

Maran1947:
@RichardAkman I've implemented the changes as per your suggestion. Could you please take a moment to review the updates and provide feedback?

ts/packages/node/src/stores/XmlFile.ts
```diff
@@ -0,0 +1,42 @@
...
+        value,
+      }));
+    });
+
+    return output;
+  }
+
+  stringify(nextConfigs: Config[]): string {
+    const builder = new xml2js.Builder();
+    const xmlObject = { root: nextConfigs }; // Assuming 'root' as the main element
```
RichardAkman, line 39: The comment is appreciated but not necessary, this logic is intrinsic to any ConfigStore that extends FileConfigStore. Please remove it 🙏🏻

RichardAkman (changes requested):
Amazing work overall @Maran1947 🥇 . The implementation works but before we can finalize this PR, I need to clarify the two following points with you:

1. In both the store class and the test file you explicitly selected the initial state of XML files to set the version to `1.0`. I'm not entirely certain why considering version `1.1` is the latest. Is there any particular reason you made this choice? 
2. The data storage strategy that was selected seemingly does not resonate with the natural structure of XML files. In other ConfigStores, configs are all grouped by some syntax that inherently exclaims that a `key`, `set`, and a `value` are connected. For example, JSON groups these in an object, and INI uses the `section` property. In this instance, the grouping is implicit via the order of each element. I'd like to understand this choice as my experience with XML files leads me in a slightly different direction. Something like this comes to mind:
```xml
  <config>
    <set>dev</set>
    <key>GREETING</key>
    <value>hey</value>
  </config>
  <config>
    <set>prod</set>
    <key>SUBJECT</key>
    <value>world</value>
  </config>
```

ts/packages/node/src/stores/XmlFile.ts
Maran1947: Alright.

Maran1947:
Thank you for your appreciation and for raising these points. 

1. Regarding the version number, there is no particular reason for using 1.0;  I will update the code to use XML version 1.1 to stay current with the latest standards..
2. Thank you for clarifying the data storage strategy. I will consider your suggestion to use explicit grouping in the XML file to better align with the natural structure of XML, improving readability and maintainability.

I will make these updates based on your feedback.

Maran1947:
@RichardAkman Done, can you please check it?

ts/packages/node/src/stores/XmlFile.ts
```diff
@@ -0,0 +1,41 @@
+import xml2js from 'xml2js';
+import { type Config } from '@configu/ts';
+import _ from 'lodash';
```
RichardAkman, line 3: lodash is now an unused import in the new implementation, please remove it 🙏🏻 
```suggestion
```

ts/packages/node/src/stores/XmlFile.ts
```diff
@@ -0,0 +1,41 @@
...
+    const parser = new xml2js.Parser({ explicitArray: false });
+    parser.parseString(fileContent, function (error, result) {
+      if (error) {
+        throw error;
+      }
+      output = result.root.config;
+    });
+
+    return output;
+  }
```
RichardAkman, line 28: The new strategy implementation is great, but a bug makes the store unusable when the file does not exist. This is because when the root tag is empty, it returns an empty string, resulting in the parse method giving a wrong response type.

image
image

To fix this, I suggest the following changes:
- Leverage the `emptyTag` property to return an empty array when no configs are under the root tag.
- if `result.root.config` is `undefined`, use the empty tag value to return an empty array.
- (Unrelated change) convert the callback to an arrow function to align with the project Eslint rules.

```suggestion
  parse(fileContent: string): Config[] {
    let output: Config[] = [];
    const parser = new xml2js.Parser({ explicitArray: false, emptyTag: () => [] });
    parser.parseString(fileContent, (error, result) => {
      if (error) {
        throw error;
      }
      output = result.root.config ?? result.root;
    });

    return output;
  }
```

ts/packages/node/src/stores/XmlFile.ts
```diff
@@ -0,0 +1,41 @@
...
+  stringify(nextConfigs: Config[]): string {
+    const options = {
+      xmldec: {
+        version: '1.1',
+        encoding: 'UTF-8',
+      },
+    };
+    const builder = new xml2js.Builder(options);
+    const xmlObject = { root: { config: nextConfigs } };
+    return builder.buildObject(xmlObject);
```
RichardAkman, line 39: I suggest that we let users of the store customize this behavior by letting them provide the builder options through the store constructor instead of picking arbitrary options.

```suggestion
  stringify(nextConfigs: Config[]): string {
    const xmlObject = { root: { config: nextConfigs } };
    return this.builder.buildObject(xmlObject);
```

ts/packages/node/src/stores/XmlFile.ts
```diff
@@ -0,0 +1,41 @@
...
+export type XmlFileConfigStoreConfiguration = { path: string };
+
+export class XmlFileConfigStore extends FileConfigStore {
+  constructor({ path }: XmlFileConfigStoreConfiguration) {
+    const initialFileState = `
+<?xml version="1.1" encoding="UTF-8"?>
+<root/>
+    `;
+    super('xml-file', { path, initialFileState });
+  }
```
RichardAkman, line 15: Let's allow users of the store to customize the builder according to their requirements.

```suggestion
export type XmlFileConfigStoreConfiguration = { path: string; builderOptions?: xml2js.BuilderOptions };

export class XmlFileConfigStore extends FileConfigStore {
  private builder: xml2js.Builder;
  constructor({ path, builderOptions }: XmlFileConfigStoreConfiguration) {
    const builder = new xml2js.Builder(builderOptions);
    const initialFileState = builder.buildObject({ root: { config: [] } });
    super('xml-file', { path, initialFileState });
    this.builder = builder;
  }
```

ts/packages/node/src/stores/XmlFile.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    ];
+    await store.set(configsToWrite);
+
+    // * Read the file again to verify the write operation
+    const newConfigs = await store.get([
+      { set: 'dev', key: 'GREETING' },
+      { set: 'prod', key: 'SUBJECT' },
+    ]);
+    expect(newConfigs).toEqual(configsToWrite);
+  });
```
RichardAkman, line 60: After catching the bug with writing to a non-existing store file. I suggest we add these tests to catch such future cases: 

```suggestion
  const writeConfigs = async (store: XmlFileConfigStore) => {
    const configsToWrite = [
      { set: 'dev', key: 'GREETING', value: 'hey' },
      { set: 'prod', key: 'SUBJECT', value: 'world' },
    ];
    await store.set(configsToWrite);

    // * Read the file again to verify the write operation
    const newConfigs = await store.get([
      { set: 'dev', key: 'GREETING' },
      { set: 'prod', key: 'SUBJECT' },
    ]);
    expect(newConfigs).toEqual(configsToWrite);
  };

  test('should write configurations to new XML file', async () => {
    await fs.unlink(testXmlFilePath);
    const store = new XmlFileConfigStore({ path: testXmlFilePath });
    await store.init();
    await writeConfigs(store);
  });

  test('should write configurations to an existing XML file', async () => {
    const store = new XmlFileConfigStore({ path: testXmlFilePath });
    await writeConfigs(store);
  });

  test('should query new XML file dev configurations', async () => {
    await fs.unlink(testXmlFilePath);
    const store = new XmlFileConfigStore({ path: testXmlFilePath });
    await store.init();
    const queries = [{ set: 'dev', key: 'GREETING' }];
    const results = await store.get(queries);
    expect(results).toEqual([]);
  });
```

RichardAkman (changes requested):
Great job on the adjustments, take a look at the issue I found and my final suggestions 🙏🏻 🎖️

Maran1947:
Updated the changes, please check it.

RichardAkman (approved):
Great work on the changes. I caught an issue with the `emptyTag` usage that resulted in the root set not working as it returned an empty array instead of an empty string. I've modified the implementation to remove the use of `emptyTag` and simply just return an empty array in case `result.root.config` is undefined. Now everything works as expected.

Thanks again for your contribution, it is highly appreciated 🌟 🌟 🌟

PR #361, opened 2024-02-13
Add --prefix and --suffix Flags to ExportCommand in CLI

### LINKED ISSUE #358: Add `--prefix` and `--suffix` Flags to ExportCommand in CLI
### Suggestion

Enhance the `ExportCommand` functionality in the CLI by introducing two new optional flags: `--prefix` and `--suffix`. These flags will allow users to append a fixed string to the beginning (`--prefix`) or end (`--suffix`) of each Config Key in the export result.

Example Usage:
```bash
conifgu eval ... | configu export --prefix "MYAPP_" --suffix "_PROD"
```
In the above example, if we have a Config Key named `DATABASE_URL`, the export result would transform it into `MYAPP_DATABASE_URL_PROD`. This functionality is particularly useful when different environments or systems require specific naming conventions for configuration keys.

### Motivation

The addition of `--prefix` and `--suffix` flags will provide users with greater control over the format of their exported configuration keys, making it easier to adapt the output for different deployment environments or system requirements.

### Context

connected to #333 

```js
    prefix: Flags.string({
      description: `Append a fixed string to the beginning of each Config Key in the export result`,
    }),
    suffix: Flags.string({
      description: `Append a fixed string to the end of each Config Key in the export result`,
    }),
```

RonConfigu:
Closes #358

### DIFF
--- a/ts/packages/cli/src/commands/export.ts
+++ b/ts/packages/cli/src/commands/export.ts
@@ -41,6 +41,10 @@ export default class Export extends BaseCommand<typeof Export> {
       description: `Pipe eval commands result to export command to pass \`Configs\` as environment variables to a child-process`,
       command: `<%= config.bin %> eval ... | <%= config.bin %> <%= command.id %> --run 'node index.js'`,
     },
+    {
+      description: `Pipe eval commands result to export command and apply a prefix / suffix to each Config Key in the export result`,
+      command: `<%= config.bin %> eval ... | configu export --prefix "MYAPP_" --suffix "_PROD"`,
+    },
   ];
 
   static flags = {
@@ -49,13 +53,11 @@ export default class Export extends BaseCommand<typeof Export> {
       default: true,
       allowNo: true,
     }),
-
     explain: Flags.boolean({
       description: `Outputs metadata on the exported \`Configs\``,
       aliases: ['report'],
       exclusive: ['format', 'template', 'source', 'run'],
     }),
-
     format: Flags.string({
       description: `Format exported \`Configs\` to common configuration formats. Redirect the output to file, if needed`,
       options: CONFIG_FORMAT_TYPE,
@@ -69,7 +71,6 @@ export default class Export extends BaseCommand<typeof Export> {
       aliases: ['EOL'],
       dependsOn: ['format'],
     }),
-
     template: Flags.string({
       description: `Path to a file containing {{mustache}} templates to render (inject/substitute) the exported \`Configs\` into`,
       exclusive: ['explain', 'format', 'source', 'run'],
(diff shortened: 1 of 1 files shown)


ts/packages/cli/src/commands/export.ts
```diff
@@ -41,6 +41,10 @@ export default class Export extends BaseCommand<typeof Export> {
       description: `Pipe eval commands result to export command to pass \`Configs\` as environment variables to a child-process`,
       command: `<%= config.bin %> eval ... | <%= config.bin %> <%= command.id %> --run 'node index.js'`,
     },
+    {
```
RichardAkman, line 44: Not entirely sure why but this addition does not show up when I run `--help`:
image

We should check if this does show up when we create a staging build.

ts/packages/cli/src/commands/export.ts
```diff
@@ -156,6 +161,12 @@ export default class Export extends BaseCommand<typeof Export> {
     this.printStdout(formattedConfigs);
   }
 
+  keysMutations() {
+    return this.flags.prefix || this.flags.suffix
+      ? (key: string) => `${this.flags.prefix ?? ''}${key}${this.flags.suffix ?? ''}`
+      : undefined;
+  }
```
RichardAkman, line 168: I like the strategy of having a builder function for `keys` but I think we can make this syntax a bit more extensible by going for something like this:
```node.js
  keysMutations() {
    return (key: string) => {
      let mutatedKey = key;
      if (this.flags.prefix) mutatedKey = `${this.flags.prefix}${mutatedKey}`;
      if (this.flags.suffix) mutatedKey = `${mutatedKey}${this.flags.suffix}`;
      return mutatedKey;
    };
  }
```

This will allow us to add any future mutation flags easily without having to refactor this code. What do you think?

RonConfigu: This will cause `ExportCommand` to **always** run the keys function. I want `keysMutations` to return undefined when no mutations needed.
How about this:
```typescript
  keysMutations() {
    if (![this.flags.prefix, this.flags.suffix].some((value) => !!value)) return undefined;
    return (key: string) => {
      let mutatedKey = key;
      if (this.flags.prefix) mutatedKey = `${this.flags.prefix}${mutatedKey}`;
      if (this.flags.suffix) mutatedKey = `${mutatedKey}${this.flags.suffix}`;
      return mutatedKey;
    };
  }
```

RichardAkman (approved):
LGTM, great work!

ts/packages/cli/src/commands/export.ts
```diff
@@ -156,6 +161,18 @@ export default class Export extends BaseCommand<typeof Export> {
     this.printStdout(formattedConfigs);
   }
 
+  keysMutations() {
+    if ([this.flags.prefix, this.flags.suffix].some((flag) => flag !== undefined)) {
```
rannn505, line 165: what if one of them is undefined?

image

rannn505:
@RonConfigu We should potenitally add validation for the prefix and suffix flags to be aligned with the key NAME patterns, as this error is a bit confusing / or should we just allow that as mutations should not be validated (fist we validate then mutate right?)
image

ts/packages/cli/src/commands/export.ts
```diff
@@ -156,6 +161,18 @@ export default class Export extends BaseCommand<typeof Export> {
...
+      return (key: string) => {
+        let mutatedKey = key;
+        if (this.flags.prefix) mutatedKey = `${this.flags.prefix}${mutatedKey}`;
+        if (this.flags.suffix) mutatedKey = `${mutatedKey}${this.flags.suffix}`;
+        return mutatedKey;
+      };
+    }
+    return undefined;
+  }
```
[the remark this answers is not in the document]
RonConfigu: Lol, https://github.com/configu/configu/pull/361#discussion_r1489272915

RonConfigu:
We're validating with NAME and then mutate. 
Should we or not.. I guess we should. it is a system-wide constraint that keys should pass the NAME test

PR #362, opened 2024-02-13
Add casing option to export command

### LINKED ISSUE #359: Add `--casing` Option to `ExportCommand`
### Suggestion

Introduce a `--casing` flag to the `ExportCommand` in the Configu CLI, enabling users to adjust the casing convention of configuration keys in the export results. The flag is optional and accepts various casing options, all presented in `pascalCase`:

- `CamelCase`: Transforms keys to `twoWords`
- `CapitalCase`: Transforms keys to `Two Words`
- `ConstantCase`: Transforms keys to `TWO_WORDS`
- `DotCase`: Transforms keys to `two.words`
- `KebabCase`: Transforms keys to `two-words`
- `NoCase`: Transforms keys to `two words`
- `PascalCase`: Transforms keys to `TwoWords`
- `PascalSnakeCase`: Transforms keys to `Two_Words`
- `PathCase`: Transforms keys to `two/words`
- `SentenceCase`: Transforms keys to `Two words`
- `SnakeCase`: Transforms keys to `two_words`
- `TrainCase`: Transforms keys to `Two-Words`

Each option will transform the casing of config keys in the output according to the selected style.

Example Usage:
```bash
configu eval ... | configu export --casing "snakeCase"
```

In the above command, if a Config Key is named `databaseURL`, applying the `--casing "snakeCase"` flag will transform it to `database_url` in the export result. This feature caters to developers and systems requiring specific key formatting, such as environment variables in UNIX systems which commonly use `UPPER_SNAKE_CASE`, or JavaScript projects that prefer `camelCase` for object keys.

This addition will not only streamline the configuration management process but also reduce manual post-processing efforts, making Configu an even more powerful tool in the hands of developers.

### Motivation

Offering the `--casing` option empowers developers to tailor the format of config keys to align with project standards or the requirements of different deployment targets. This capability simplifies configuration management and deployment processes, especially when working across diverse platforms.

### Context

The `change-case` library, a part of the project's dependencies, provides the functionality necessary for implementing the casing transformations.
https://www.npmjs.com/package/change-case

```js
    casing: Flags.string({
      description: `Transforms the casing of Config Keys in the export result to camelCase, PascalCase, Capital Case, snake_case, param-case, CONSTANT_CASE and others`,
    }),
```

RonConfigu:
Closes #359

### DIFF
--- a/ts/packages/cli/src/commands/export.ts
+++ b/ts/packages/cli/src/commands/export.ts
@@ -1,10 +1,21 @@
 import { cwd } from 'process';
 import { spawnSync } from 'child_process';
 import { Flags, ux } from '@oclif/core';
-import _ from 'lodash';
+import _, { kebabCase } from 'lodash';
 import { TMPL, type EvalCommandReturn, EvaluatedConfigOrigin, type ExportCommandReturn } from '@configu/ts';
 import { ExportCommand } from '@configu/node';
 import { CONFIG_FORMAT_TYPE, formatConfigs, type ConfigFormat } from '@configu/lib';
+import {
+  dotCase,
+  constantCase,
+  noCase,
+  pascalCase,
+  pathCase,
+  sentenceCase,
+  snakeCase,
+  capitalCase,
+  camelCase,
+} from 'change-case';
 import { BaseCommand } from '../base';
 import { readFile } from '../helpers';
 
@@ -12,6 +23,21 @@ export const NO_CONFIGS_WARNING_TEXT = 'no configuration was fetched';
 export const CONFIG_EXPORT_RUN_DEFAULT_ERROR_TEXT = 'could not export configurations';
 
 type TemplateContext = { [key: string]: string } | { key: string; value: string }[];
+
+const CASE_FUNCTION: { [casing: string]: (str: string) => string } = {
+  CamelCase: camelCase,
+  CapitalCase: capitalCase,
+  ConstantCase: constantCase,
+  DotCase: dotCase,
+  KebabCase: kebabCase,
+  NoCase: noCase,
+  PascalCase: pascalCase,
+  PascalSnakeCase: (str: string) => capitalCase(str).split(' ').join('_'),
+  PathCase: pathCase,
+  SentenceCase: sentenceCase,
+  SnakeCase: snakeCase,
+  TrainCase: (str: string) => capitalCase(str).split(' ').join('-'),
+};
 
 export default class Export extends BaseCommand<typeof Export> {
   static description = `Export \`Configs\` as configuration data in various modes`;
@@ -95,6 +121,10 @@ export default class Export extends BaseCommand<typeof Export> {
     suffix: Flags.string({
       description: `Append a fixed string to the end of each Config Key in the export result`,
     }),
+    casing: Flags.string({
+      description: `Transforms the casing of Config Keys in the export result to camelCase, PascalCase, Capital Case, snake_case, param-case, CONSTANT_CASE and others`,
+      options: Object.keys(CASE_FUNCTION),
+    }),
   };
 
   printStdout(finalConfigData: string) {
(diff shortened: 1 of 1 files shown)


ts/packages/cli/src/commands/export.ts
```diff
@@ -1,10 +1,21 @@
 import { cwd } from 'process';
 import { spawnSync } from 'child_process';
 import { Flags, ux } from '@oclif/core';
-import _ from 'lodash';
+import _, { kebabCase } from 'lodash';
```
RichardAkman, line 4: Instead of using lodash, let's keep using `change-case` for this feature. For kebab case, you can use `paramCase` from `change-case`. They recently released a new major version where they renamed paramCase to kebabCase but I don't want you to have to deal with potential breaking changes.

ts/packages/cli/src/commands/export.ts
```diff
@@ -13,6 +24,21 @@ export const CONFIG_EXPORT_RUN_DEFAULT_ERROR_TEXT = 'could not export configurat
...
+  KebabCase: kebabCase,
+  NoCase: noCase,
+  PascalCase: pascalCase,
+  PascalSnakeCase: (str: string) => capitalCase(str).split(' ').join('_'),
+  PathCase: pathCase,
+  SentenceCase: sentenceCase,
+  SnakeCase: snakeCase,
+  TrainCase: (str: string) => capitalCase(str).split(' ').join('-'),
+};
+
```
RichardAkman, line 41: NIT:
- VSCode doesn't like `str` as a parameter name due to the inbuilt spellchecker which causes unnecessary warnings, `string` would be preferable.
- I think this `Record<string, (string: string) => string>` is cleaner than `{ [casing: string]: (str: string) => string }`

ts/packages/cli/src/commands/export.ts
```diff
@@ -13,6 +24,21 @@ export const CONFIG_EXPORT_RUN_DEFAULT_ERROR_TEXT = 'could not export configurat
 
 type TemplateContext = { [key: string]: string } | { key: string; value: string }[];
 
+const CASE_FUNCTION: { [casing: string]: (str: string) => string } = {
```
RichardAkman, line 27: `CASE_FUNCTION` -> `casingFormatters`, this change will align with the existing `configFormatters` from `lib` in terms of the naming pattern.

ts/packages/cli/src/commands/export.ts
```diff
@@ -41,6 +67,10 @@ export default class Export extends BaseCommand<typeof Export> {
       description: `Pipe eval commands result to export command to pass \`Configs\` as environment variables to a child-process`,
       command: `<%= config.bin %> eval ... | <%= config.bin %> <%= command.id %> --run 'node index.js'`,
     },
+    {
+      description: `Pipe eval commands result to export command and apply a prefix / suffix to each Config Key in the export result`,
+      command: `<%= config.bin %> eval ... | configu export --prefix "MYAPP_" --suffix "_PROD"`,
+    },
```
RichardAkman, line 73: looks like a mistake, should be a casing example and not prefix/suffix

ts/packages/cli/src/commands/export.ts
RonConfigu: Cool. thanks

ts/packages/cli/src/commands/export.ts
RonConfigu: Sure

ts/packages/cli/src/commands/export.ts
RonConfigu: It is. Thanks.
Changed to `eval ... | configu export --casing "snakeCase"`

ts/packages/cli/src/commands/export.ts
RichardAkman: It is a const, just for the sake of clarity:
image

RichardAkman (approved):
LGTM, great work!

ts/packages/cli/src/commands/export.ts
```diff
@@ -156,6 +186,11 @@ export default class Export extends BaseCommand<typeof Export> {
     this.printStdout(formattedConfigs);
   }
 
+  applyCasing(result: { [key: string]: string }) {
```
rannn505, line 189: this should be with the same keyMutations function of the --prefix PR

RonConfigu: Done. Thanks

RonConfigu:
@pelegpor 
> invalid config key at ExportCommand:.ConfigKey:EMPTY, key "EMPTY" mustn't contain reserved words

PR #378, opened 2024-02-21
Implement ExportCommand filter Callback

### LINKED ISSUE #368: ExportCommand `filter` Callback
### Suggestion

Introduce a `filter` callback function to the `ExportCommand`. This function accepts a single `EvalCommandReturn` object as its parameter and returns a boolean to determine whether a specific configuration key-value pair should be included in the `ExportCommandReturn` dictionary. This mechanism provides the foundation for implementing CLI flags such as `--omit-label`, `--pick-label`, `--omit-empty`, and `--include-hidden`, enabling users to finely tune the output of their exported configurations.

image

### Motivation

The introduction of a filter callback in the `ExportCommand` significantly enhances the flexibility and control users have over the configuration data they export. This feature addresses the need to selectively process and refine configuration data before it is used or deployed, catering to various operational and environmental requirements. By enabling the dynamic inclusion or exclusion of specific configurations based on custom criteria, this feature streamlines the management of configuration data across different stages of the software development lifecycle.

### Context

We should unify all filtering ops including the current hidden one to be handled in a single method

RonConfigu:
Closes #368

### DIFF
--- a/ts/packages/ts/src/commands/ExportCommand.ts
+++ b/ts/packages/ts/src/commands/ExportCommand.ts
@@ -10,6 +10,7 @@ export type ExportCommandReturn = {
 export type ExportCommandParameters = {
   pipe: EvalCommandReturn;
   keys?: (key: string) => string;
+  filter?: ({ context, result }: EvalCommandReturn['string']) => boolean;
 };
 
 export class ExportCommand extends Command<ExportCommandReturn> {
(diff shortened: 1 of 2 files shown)


ts/packages/ts/src/commands/ExportCommand.ts
```diff
@@ -37,10 +38,22 @@ export class ExportCommand extends Command<ExportCommandReturn> {
...
+
   async run() {
     const { pipe } = this.parameters;
-    const configDict = _.mapValues(pipe, (current) => current.result.value);
-    const keyMutatedConfigDict = this.mutateKeys(configDict);
-    return keyMutatedConfigDict;
+    const filteredPipe = this.filterPipe(pipe);
+    const mappedPipe = this.mapPipe(filteredPipe);
+    const keyMutatedMappedPipe = this.mutateKeys(mappedPipe);
+    return keyMutatedMappedPipe;
```
RichardAkman, line 57: I don't see the value of the `mapPipe` method, unless there's a reason that I'm missing I think we should keep the code as is:
```suggestion
    const filteredPipe = this.filterPipe(pipe);
    const configDict = _.mapValues(filteredPipe, (current) => current.result.value);
    const keyMutatedConfigDict = this.mutateKeys(configDict);
    return keyMutatedConfigDict;
```

RichardAkman:
This PR is included in #379, decided to close this one to streamline the CR process, the other PR has been linked to both issues.

PR #366, opened 2024-02-19
Implement hidden Property for Cfgu Declaration

RonConfigu:
Closes #365

PR #379, opened 2024-02-21
Implement `hidden` property for Cfgu declaration with ExportCommand `filter` callback

RonConfigu:
Closes #366


ts/packages/ts/src/commands/ExportCommand.ts
```diff
@@ -37,10 +38,23 @@ export class ExportCommand extends Command<ExportCommandReturn> {
     });
   }
 
+  private filterPipe(pipe: EvalCommandReturn) {
+    const { filter } = this.parameters;
+    if (!filter) {
+      return _.pickBy(pipe, ({ context }) => !context.cfgu.hidden);
+    }
+    return _.pickBy(pipe, filter);
+  }
```
RichardAkman, line 47: This logic feels inaccurate, if a filter is provided then the `hidden` filter is skipped and causes all hidden keys to be shown. I think the hidden filter should always happen on top of the provided filter.

ts/packages/ts/src/commands/ExportCommand.ts
```diff
@@ -37,10 +38,23 @@ export class ExportCommand extends Command<ExportCommandReturn> {
...
+
   async run() {
     const { pipe } = this.parameters;
-    const configDict = _.mapValues(pipe, (current) => current.result.value);
-    const keyMutatedConfigDict = this.mutateKeys(configDict);
-    return keyMutatedConfigDict;
+    const filteredPipe = this.filterPipe(pipe);
+    const mappedPipe = this.mapPipe(filteredPipe);
+    const keyMutatedMappedPipe = this.mutateKeys(mappedPipe);
+    return keyMutatedMappedPipe;
```
RichardAkman, line 58: I don't see the value of the `mapPipe` method, unless there's a reason that I'm missing I think we should keep the code as is:
```suggestion
    const filteredPipe = this.filterPipe(pipe);
    const configDict = _.mapValues(filteredPipe, (current) => current.result.value);
    const keyMutatedConfigDict = this.mutateKeys(configDict);
    return keyMutatedConfigDict;
```

ts/packages/ts/src/commands/commands.test.ts
```diff
@@ -377,24 +377,49 @@ describe(`commands`, () => {
     });
   });
   describe(`ExportCommand`, () => {
```
RichardAkman, line 379: NIT: Depending on your input regarding `filter` overriding `hidden` filtering, maybe we should add a test that covers a scenario where both situations occur at the same time to ensure we don't miss this situation in the future.

RichardAkman (changes requested):
Great work  ⭐

Please take a look at my comments 🙏🏻

ts/packages/ts/src/commands/ExportCommand.ts
```diff
@@ -37,9 +38,22 @@ export class ExportCommand extends Command<ExportCommandReturn> {
...
+    const { filter } = this.parameters;
+    if (!filter) {
+      return _.pickBy(pipe, ({ context }) => !context.cfgu.hidden);
+    }
+    return _.pickBy(pipe, filter);
+  }
+
+  private mapPipe(pipe: EvalCommandReturn) {
+    return _.mapValues(pipe, (current) => current.result.value);
+  }
```
RichardAkman, line 51: I think you mistakingly left this unused method, I'll remove it for you so we can approve and finalize this PR.

RichardAkman (approved):
LGTM, great work! 🚀 ⭐

PR #381, opened 2024-02-22
Implement enhance export command with advanced filtering

### LINKED ISSUE #369: Enhance `ExportCommand` with Advanced Filtering Flags
### Suggestion

Expand the `ExportCommand` in the Configu CLI to include comprehensive filtering flags, offering users refined control over their configuration export results. These new flags will invoke the filter callback functionality, allowing for more granular selection or exclusion of configuration data based on labels, key names, visibility, and value presence.

- `--pick-label` / `--omit-label`: Specify labels to include or exclude. Can be used multiple times to specify multiple labels.
  
- `--pick-key` / `--omit-key`: Specify exact config keys to include or exclude. Can be used multiple times for multiple keys. this is casing sensitive and according to the original key specified from the user's schema, the key mutations come last.
  
- `--pick-hidden` / `--omit-hidden`: Explicitly include or exclude config keys marked as hidden. By default, hidden keys are omitted.
  
- `--pick-empty` / `--omit-empty`: Choose to include or exclude config keys with empty values. By default, empty values are included.

**Important notes**
- The order of the flags sets the priority of the filter when there are collisions
- If any `pick` flag is specified all other are omitted by default
- if any `omit` flag is specified all other are included by default
- if a combination of pick and omit are specified, keep the above. for example `--pick-label 'aws' --omit-key 'AWS_SECRET_ACCESS_KEY'` will return all `aws` labeled keys without the secret access key.

**Example Usage:**
```bash
# Exclude specific labels
configu eval ... | configu export --omit-label 'deprecated' --omit-label 'temporary'

# Include only configs under specific labels
configu eval ... |  configu export --pick-label 'production' --pick-label 'secure'

# Exclude specific keys
configu eval ... | configu export --omit-key 'DEBUG_MODE' --omit-key  'TEST_ACCOUNT'

# Include hidden configs and Exclude configs with empty values
configu eval ... | configu export --pick-hidden  --omit-empty
```

### Motivation

These enhancements aim to provide users with intuitive, command-line-driven mechanisms to tailor their configuration export results precisely, catering to diverse operational and deployment scenarios. By offering the ability to include or exclude configurations based on their characteristics, users gain more flexibility in managing configurations for different environments or purposes, aligning with best practices in configuration management.

### Context

This feature builds on the recently introduced `filter` callback functionality within the `ExportCommand`, translating it into practical, user-friendly CLI options.

- Connected to #368 #367 
- you can know the order of the flags from oclif API by extending the base command class:
image

RonConfigu:
Closes #369


ts/packages/cli/src/commands/export.ts
```diff
@@ -130,6 +146,38 @@ export default class Export extends BaseCommand<typeof Export> {
...
+      exclusive: ['pick-hidden'],
+    }),
+    'pick-empty': Flags.boolean({
+      description: `Include config keys with empty values. By default, empty values are included.`,
+      exclusive: ['omit-empty'],
+    }),
+    'omit-empty': Flags.boolean({
+      description: `Exclude config keys with empty values. By default, empty values are included.`,
+      exclusive: ['pick-empty'],
+    }),
```
RichardAkman, line 180: Please remove the period at the end of the descriptions to align with the rest of the example descriptions. `format` for example has a period in the middle of the sentence but not at the end.

ts/packages/cli/src/base.ts
```diff
@@ -198,14 +200,15 @@ export abstract class BaseCommand<T extends typeof Command> extends Command {
...
-    const { args, flags } = await this.parse({
+    const { args, flags, raw } = await this.parse({
       flags: this.ctor.flags,
       baseFlags: (super.ctor as typeof BaseCommand).baseFlags,
       args: this.ctor.args,
       strict: this.ctor.strict,
     });
     this.flags = flags as Flags<T>;
     this.args = args as Args<T>;
+    this.rawFlags = raw as FlagToken[];
```
RichardAkman, line 211: I need some clarification regarding the decision to use the `FlagToken` type. The type of `raw` is `ParsingToken[]` and casting it to `FlagToken[]` seems dangerous. Is there something that guarantees that the tokens aren't of type `ArgToken`?

RonConfigu: There are no ArgToken for export command.
This is what will happen if you try to use one
image

ts/packages/cli/src/commands/export.ts
```diff
@@ -213,26 +267,67 @@ export default class Export extends BaseCommand<typeof Export> {
     return caseFunction ? _.mapKeys(result, (value, key) => caseFunction(key)) : result;
   }
 
+  filterFromFlags(): (({ context, result }: EvalCommandReturn['string']) => boolean) | undefined {
+    const filterFlags = _.pickBy(this.flags, (value, key) => Object.values(FilterFlag).includes(key as FilterFlag));
+    if (!_.isEmpty(filterFlags)) {
```
RichardAkman, line 272: NIT: I prefer we go for the early return pattern and do:
```node.js
 if (_.isEmpty(filterFlags)) {
    return undefined;
  }
```

This will make the code more readable and reduce indenting

ts/packages/cli/src/commands/export.ts
```diff
@@ -213,26 +267,67 @@ export default class Export extends BaseCommand<typeof Export> {
...
+      const pickFlags = _.pick(filterFlags, [FilterFlag.PickLabel, FilterFlag.PickKey]);
+      const omitFlags = _.pick(filterFlags, [FilterFlag.OmitLabel, FilterFlag.OmitKey]);
+      return ({ context, result }: EvalCommandReturn['string']): boolean => {
```
RichardAkman, line 275: NIT: the implementation is solid but the usage of reducers makes this code verbose and hard to read. How do you feel about this suggestion:
```node.js
    const pickedKeys = (_.pick(filterFlags, [FilterFlag.PickKey])['pick-key'] as string[]) ?? [];
    const pickedLabels = (_.pick(filterFlags, [FilterFlag.PickLabel])['pick-label'] as string[]) ?? [];
    const omittedKeys = (_.pick(filterFlags, [FilterFlag.OmitKey])['omit-key'] as string[]) ?? [];
    const omittedLabels = (_.pick(filterFlags, [FilterFlag.OmitLabel])['omit-label'] as string[]) ?? [];

    return ({ context, result }: EvalCommandReturn['string']): boolean => {
      const hiddenFilter = this.flags[FilterFlag.PickHidden] ? true : !context.cfgu.hidden;
      const emptyFilter = this.flags[FilterFlag.OmitEmpty] ? result.origin !== EvaluatedConfigOrigin.EmptyValue : true;

      const isKeyPicked = pickedKeys.includes(context.key);
      const isLabelPicked = _.intersection(pickedLabels, context.cfgu.labels ?? []).length > 0;
      const pickFilter = _.isEmpty([...pickedKeys, ...pickedLabels]) || isKeyPicked || isLabelPicked;

      const isKeyOmitted = omittedKeys.includes(context.key);
      const isLabelOmitted = _.intersection(omittedLabels, context.cfgu.labels ?? []).length > 0;
      const omitFilter = !isKeyOmitted && !isLabelOmitted;

      return hiddenFilter && emptyFilter && pickFilter && omitFilter;
    };
```

Feel free to take whichever part of it you like if any.

RonConfigu: I prefer to use vanilla instead of lodash every day.
The only reason I use it here is because eslint doesn't allow to reduce but misses _.reduce.
For me, it's harder to read your code than mine... :/

But, I'm not in love with mine. If it matters to you i'll use yours.

RichardAkman (approved):
LGTM, great work 🌟 🌟 🌟

PR #385, opened 2024-03-06
Fix `LocalForageConfigStore` storing stringified data

### LINKED ISSUE #384: Fix `LocalForageConfigStore` storing stringified data
### Current Behavior

Any configs upserted through `LocalForageConfigStore` are stored stringified which is unintentional behavior.

### Expected Behavior

The configs should be stored in their natural way unstringified.

### To Reproduce

Perform an upsert with any configs using the store and view the browser cache to see the value being stored in a `JSON.stringify` format.

### Context

Localforage internally performs `JSON.stringify` on any data before being stored and this causes the issue because `KeyValueConfigStore` already performs `JSON.stringify` internally.

RonConfigu:
- getByKey is now aware of null and not stringifying it
- upsert will try to parse the values

Closes #384


ts/packages/cli/src/base.ts
```diff
@@ -242,6 +242,8 @@ export abstract class BaseCommand<T extends typeof Command> extends Command {
       });
       const cliConfigData = JSON.parse(compiledCliConfigData);
       this.config.cli.data = cliConfigData;
+      this.log(this.config.configu.file);
+      this.log(`cli config: ${JSON.stringify(this.config.cli)}`);
```
RichardAkman, line 246: Please remove

RichardAkman (approved):
LGTM ⭐

PR #389, opened 2024-03-07
 Implement SDK changes for local cache for config store

### LINKED ISSUE #370: Implement Local Cache for ConfigStore
### Suggestion
We propose the integration of a local caching mechanism for ConfigStore which will significantly enhance performance by reducing fetch times and providing a fallback during outages. This cache should be configurable via the `.configu` file, allowing users to toggle caching on or off, or specify a custom SQLite file for the cache data.

- Add an optional string `cache` property to root of the `.configu` file.
- Add an optional boolean `cache` property under `stores.[storeName].cache` of the `.configu` file.
- Enable local cache if one of the `stores.[storeName].cache` props are true. -> this enables the cache for that spesific store. 
- If there arent any stores with cache: true, disable cache.
- if a custom path string is provided to the root cache prop -> use the specified SQLite file otherwise use the default cache path at `$cacheDir/configs.sqlite`:
- Automatically save to cache after each successful `eval` command execution involving that ConfigStore.
- In case of a ConfigStore fetch error during `eval` (not an actual eval error like validate the cfgu against a Config), attempt to fetch the Config from the local cache if caching is enabled.
- Add a `--force-cache` flag to the `eval` CLI command to allow fetching Configs strictly from the cache.

### Motivation
Introducing a local cache for ConfigStore aims to:
- Minimize latency during configuration fetch operations.
- Provide resilience against temporary ConfigStore outages or network issues.
- Streamline CI/CD processes by allowing cache artifacts to be saved and reused, improving build times and reliability.

### Context
Additional considerations include:
- Ensuring cache integrity and freshness—cache should be updated or invalidated based on configurable criteria.
- Testing within CI environments like GitHub Actions to confirm the cache's effectiveness as a stored artifact and its reusability across multiple build runs.
- Establishing a clear cache management strategy for users to understand when and how cache is utilized or ignored.

RonConfigu:
> ORM sets the name for the `set:INDEX` automatically to avoid naming conflicts

Closes #370


ts/packages/ts/src/commands/EvalCommand.ts
```diff
@@ -245,7 +250,11 @@ export class EvalCommand extends Command<EvalCommandReturn> {
             ConfigSchema.CFGU.VALIDATORS.valueOptions(cfgu, evaluatedValue);
             ConfigSchema.CFGU.VALIDATORS.valueType(cfgu, evaluatedValue);
           } catch (error) {
-            throw error?.appendScope?.(errorScope) ?? error;
+            let err = error?.appendScope?.(errorScope) ?? error;
+            if (current.result.origin === EvaluatedConfigOrigin.StoreSet) {
+              err = new ConfigStoreError('invalid config value', err.message, err?.scope ?? errorScope);
+            }
```
RichardAkman, line 256: In my opinion, this shouldn't result in a store error. Let's get product opinion on this.

ts/packages/ts/src/commands/EvalCommand.ts
```diff
@@ -60,37 +60,42 @@ export class EvalCommand extends Command<EvalCommandReturn> {
...
+          ...current,
+          result: {
+            origin: EvaluatedConfigOrigin.StoreSet,
+            source: `parameters.store=${context.store},parameters.set=${context.set}`,
+            value: storeConfig.value,
+          },
+        };
+      });
+    } catch (error) {
+      throw new ConfigStoreError('failed to get store configs', error.message, [['EvalCommand', 'evalFromStoreSet']]);
```
RichardAkman, line 97: I'm unsure why everything is wrapped in a try-catch instead of `store.get` exclusively. Please explain the rationale.

ts/packages/node/src/stores/ORM.ts
```diff
@@ -1,59 +1,72 @@
...
   }
 
   private async upsert(configs: IConfig[]): Promise<void> {
-    const configRepository = this.dataSource.getRepository(Config);
+    const configRepository = this.dataSource.getRepository(this.table);
 
     if (configs.length > 0) {
       await configRepository.upsert(configs, ['set', 'key']);
     }
+    // await this.dataSource.destroy();
```
RichardAkman, line 65: Please remove unnecessary code 🙏🏻

ts/packages/node/src/stores/SQLite.ts
```diff
@@ -1,7 +1,9 @@
 import { type SqliteConnectionOptions } from 'typeorm/driver/sqlite/SqliteConnectionOptions';
 import { ORMConfigStore } from './ORM';
 
-export type SQLiteConfigStoreConfiguration = Omit<SqliteConnectionOptions, 'type'>;
+export type SQLiteConfigStoreConfiguration = Omit<SqliteConnectionOptions, 'type'> & {
+  tableName?: string;
+};
```
RichardAkman, line 6: I feel like modifying `ORMConfigStore` to accept a table only for `SQLiteConfigStore` is quirky. I feel like we should make this new property available for all stores extending `ORMConfigStore`. Let's get a product opinion on this.

ts/packages/node/src/stores/ORM.ts
```diff
@@ -1,59 +1,72 @@
...
+  return Config;
+};
+
+type ORMConfigStoreOptions = DataSourceOptions & {
+  tableName?: string;
+};
 
 export abstract class ORMConfigStore extends ConfigStore {
   readonly dataSource: DataSource;
+  private readonly table: ReturnType<typeof createEntity>;
```
RichardAkman, line 33: I think the disconnect between the class property name and the parameter `tableName` is a source of confusion. I suggest that we align the naming either way.

ts/packages/node/src/stores/ORM.ts
```diff
@@ -1,59 +1,72 @@
...
+  @Entity({ name: tableName })
+  @Index(['set', 'key'], { unique: true })
+  class Config {
+    @PrimaryGeneratedColumn('uuid')
+    id: string;
 
-  @Index('set')
-  @Column('text')
-  set: string;
+    @Index(`${tableName}_set`)
```
RichardAkman, line 13: I'm unsure about the meaning of this change, please explain.

ts/packages/node/src/stores/ORM.ts
RichardAkman, line 33: I suggest `configEntity` instead of `table` to keep things clearer.

ts/packages/node/src/stores/ORM.ts
RonConfigu: Index name is unique across the entire db. it has nothing to do with the actual indexing

RichardAkman: Each table should have an index on the set column, why should the index name be unique? There's no restriction that I'm aware of.

RichardAkman: As discussed, let's just give ORM the responsibility for naming the index. Please mention this as a note on the PR for future reference.

ts/packages/ts/src/commands/EvalCommand.ts
RichardAkman: Ran agrees

ts/packages/node/src/stores/SQLite.ts
RichardAkman: All ORM config stores should have the new table name props

RonConfigu: Done

ts/packages/ts/src/commands/EvalCommand.ts
RonConfigu: Revert

RichardAkman (approved):
LGTM, great work 🎖️

PR #392, opened 2024-03-11
Fix CLI casing functions to remove white-spaces

RonConfigu:
Closes #


RichardAkman:
After a discussion with @rannn505, the solution applied in this PR doesn't align with the product vision. I'll open an issue and Ran will decide on a solution.

PR #394, opened 2024-03-11
Implement CSV ConfigStore in Node.js SDK

### LINKED ISSUE #151: Implement `CSV` `ConfigStore` in Node.js SDK

Maran1947:
Closes #151 
Hi @RichardAkman, Can you please take a moment to review it?

### DIFF
--- a/ts/packages/node/src/stores/CsvFile.ts
+++ b/ts/packages/node/src/stores/CsvFile.ts
@@ -1,0 +1,23 @@
+import { type Config } from '@configu/ts';
+import { parse, stringify } from 'csv/sync';
+import { FileConfigStore } from './File';
+
+export type CsvFileConfigStoreConfiguration = { path: string };
+
+export class CsvFileConfigStore extends FileConfigStore {
+  constructor({ path }: CsvFileConfigStoreConfiguration) {
+    const initialFileState = '';
+    super('csv-file', { path, initialFileState });
+  }
+
+  parse(fileContent: string): Config[] {
+    const csvRows = parse(fileContent);
+    const configs = csvRows.slice(1).map(([set, key, value]: [string, string, string]) => ({ set, key, value }));
+    return configs;
+  }
+
+  stringify(nextConfigs: Config[]): string {
+    const csvRows = [['set', 'key', 'value'], ...nextConfigs.map(({ set, key, value }) => [set, key, value])];
+    return stringify(csvRows);
+  }
+}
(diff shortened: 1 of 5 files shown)


ts/packages/node/src/stores/CsvFile.ts
```diff
@@ -0,0 +1,23 @@
...
+  parse(fileContent: string): Config[] {
+    const csvRows = parse(fileContent);
+    const configs = csvRows.slice(1).map(([set, key, value]: [string, string, string]) => ({ set, key, value }));
+    return configs;
+  }
+
+  stringify(nextConfigs: Config[]): string {
+    const csvRows = [['set', 'key', 'value'], ...nextConfigs.map(({ set, key, value }) => [set, key, value])];
+    return stringify(csvRows);
+  }
```
RichardAkman, line 22: We should utilize the capabilities of the `csv` package to streamline these methods:

```suggestion
  parse(fileContent: string): Config[] {
    return parse(fileContent, { columns: true });
  }

  stringify(nextConfigs: Config[]): string {
    return stringify(nextConfigs, { header: true });
  }
```

RichardAkman (changes requested):
Great work, your experience implementing config stores is really showing ⭐ 

Take a look at my suggestion and then we can finalize this PR 🚀

Maran1947:
Thanks for appreciation. I've really enjoyed working on config stores, and your support means a lot to me! 🙌 🙏

I've noticed that the `parse` method breaks when encountering empty lines, so I've added a `skip_empty_lines` attribute to handle this issue. Additionally, I've included a `trim` attribute to handle spaces and tabs. 

Could you please review these changes and suggest any improvements?

RichardAkman (approved):
I've removed the addition of `trim` as we do not want to mutate the data that the users of the config store are storing. Looks great now, the addition of `skip_empty_lines` is solid. 🏅

Issue #377, opened 2024-02-21
Format objects for Terraform .tfvars file


samgabrail:
### Current Behavior

using a schema like this:
```json
"my_tags": {
"type": "JSONSchema",
"description": "A map of tags to add to all resources.",
"schema": {"type": "object"},
"required": true
}
```
then exporting as a Terraform .tfvars file results in this:
```bash
prefix = "configu"
region = "us-east-1"
address_space = "10.0.0.0/16"
subnet_prefix = "10.0.10.0/24"
instance_type = "t2.micro"
my_aws_key = "mykey.pem"
my_tags = "{"Name": "configu-demo", "Environment": "Development"}"
```

### Expected Behavior

However, Terraform doesn't like this and considers it invalid. would need to either do this:
`my_tags = "{\"Name\": \"configu-demo\", \"Environment\": \"Development\"}"`

or better yet, export it as a map (object) like this:
`my_tags = {"Name": "configu-demo", "Environment": "Development"}`

RichardAkman:
Great catch @samgabrail!

@rannn505 please provide your insight regarding the right way to go about fixing this issue.

PR #395, opened 2024-03-13
Format objects for terraform .tfvars file 

idanela:
When coverting to .tfvars files the code ignored the possibility that value could be an object.
Therefore, terraform did not considered it valid because it was converted to a string that contains the object.
I added a check to determine rather it is an object or not, and if so it ommited the quotation marks.

Solves #377


rannn505:
@RichardAkman can you review?

RichardAkman (approved):
Great work ⭐ 🏅 

I've made a slight adjustment to the formatter strategy to ensure that the function always correctly identifies object strings.

PR #412, opened 2024-04-15
Add TODO comment to `KeyvConfigStore`

RichardAkman:
After a discussion with @rannn505, we agreed that `KeyvConfigStore` should be reworked to make it usable by the CLI. Until we decide to invest time for it, we agreed to add a TODO comment in the store implementation file.


RonConfigu (approved):
Superb

PR #413, opened 2024-04-15
Refactor Local Cache for ConfigStore SDK

### LINKED ISSUE #370: Implement Local Cache for ConfigStore

RonConfigu:
Closes #370

### DIFF
--- a/ts/packages/ts/src/commands/EvalCommand.ts
+++ b/ts/packages/ts/src/commands/EvalCommand.ts
@@ -1,7 +1,7 @@
 import _ from 'lodash';
 import { Command } from '../Command';
 import { type Cfgu } from '../types';
-import { ConfigError, TMPL, ConfigStoreError } from '../utils';
+import { ConfigError, TMPL } from '../utils';
 import { type ConfigStore } from '../ConfigStore';
 import { ConfigSet } from '../ConfigSet';
 import { ConfigSchema } from '../ConfigSchema';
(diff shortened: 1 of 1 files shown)


RichardAkman (approved):
LGTM

PR #415, opened 2024-04-16
Fix cache store throws exception if storeFlag is missing

### LINKED ISSUE #414: Cache store throws exception if storeFlag is missing
### Current Behavior

EvalCommand can run withour a store to eval schema.
With the new Cache mech this is not working.

### Expected Behavior

No caching for NoopStore

### To Reproduce

_No response_

### Context

_No response_

RonConfigu:
Closes #414


RichardAkman (approved):
LGTM

PR #442, opened 2024-08-08
using another OSS project for login

davidmeirlevy:
Closes #


rannn505:
@davidmeirlevy we will reopen it when we done with Zitadel on @configu/platform

PR #447, opened 2024-09-15
#446 - Ability to register integrations into configu

### LINKED ISSUE #446: Ability to register integrations into configu
### Suggestion

a user would like to use custom behavior for:
- schema validations and manipulations (expressions)
- stores
- exporters (to different types of structures).

### Motivation

we as Cofigu would also like to have a separate behavior to store our internal mechanisms and re-use them in both CLI and proxy service.

### Context

_No response_

davidmeirlevy:
Closes #446

### DIFF
--- a/.github/workflows/cd-configu-integrations.yml
+++ b/.github/workflows/cd-configu-integrations.yml
@@ -1,0 +1,83 @@
+name: Build Deferred Integrations
+
+on:
+  push:
+    branches:
+      - 446-ability-to-register-integrations-into-configu
+      - main
+
+env:
+  PACKAGE_NAME: 'configu-integrations'
+  PACKAGE_PATH: 'packages/integrations'
+  VERSION: 'latest'
+
+jobs:
+  #  start-release:
+  #    runs-on: ubuntu-latest
+  #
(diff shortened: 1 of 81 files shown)


davidmeirlevy:
example to lazy load of store:

```
static async getOne(key: string) {
    const store = allStores.get(key);
    if (store) {
      return store;
    }
    const VERSION = 'latest';
    const CONFIGU_HOME = path.join(process.cwd(), '/.configu-cache');

    const STORE_PATH = path.join(CONFIGU_HOME, `/${key}-${VERSION}.js`);

    if (!existsSync(STORE_PATH)) {
      const res = await fetch(
        `https://github.com/configu/configu/releases/download/${VERSION}/${key}-${platform()}.js`,
      );
      if (res.ok) {
        await mkdir(CONFIGU_HOME, { recursive: true });
        await writeFile(STORE_PATH, await res.text());
      }
    }

    // eslint-disable-next-line global-require,import/no-dynamic-require
    require(STORE_PATH);
    return allStores.get(key);
  }
```

rannn505 (approved):

PR #448, opened 2024-09-15
@configu/common init

RichardAkman:
This PR creates a package that contains shared logic between the CLI and the Proxy. Includes extractions of shared logic from both packages.

## Changes
- Extract #370 from CLI into `@configu/common` by creating an `EvalCommand` that extends the base command from `@configu/ts` that implements the caching feature as `configu backup`. The feature renaming was requested by Configu product requirements. 
- Extract shared `.configu` file logic to a class interface called `ConfiguFile` (naming TBD) which includes
  - Loading `.configu` content via path to a file.
  - Loading `.configu` via the search mechanism from [cosmicconfig](https://github.com/cosmiconfig/cosmiconfig?tab=readme-ov-file#explorersearch) 
  - Running a script defined in `.configu`

## TODO
- [ ] Replace the `@configu/ts` dependency with `@configu/sdk` from #445 
- [ ] Replace the `@configu/node` dependency with `@configu/integrations` from #447 
- [ ] Retire unused packaged from `@configu/proxy` and `@configu/cli` after the above


RichardAkman:
Retired in favor of #488

PR #464, opened 2024-09-26
fix hacktoberfest logo

davidmeirlevy:
GitHub raw URLs have been changed since last year, so the filename is not the only change that should have been done.


rannn505 (approved):
Thanks @davidmeirlevy

rannn505 (approved):
Thanks @davidmeirlevy

PR #470, opened 2024-10-02
Add --configOptions flag to CLI export command 

### LINKED ISSUE #453: Merge `CompactJSON` formatter with `JSON` formatter
### Suggestion

Currently, there are two distinct versions of the `JSON` formatter. Instead, we should merge both versions into a single `JSON` formatter which by default prints the "compact" version and we should extend the formatter interface to accept parameters that modify the output to support the previous `JSON` formatter that printed the output in a "pretty" variation.

### Motivation

- There shouldn't be duplications of formatters, this is a source of confusion for users.
- The default format of formatters should be as unopinionated as possible, "prettified" outputs are inherently opinionated.
- Creating an extensible interface for formatters behavior modifications allows the user to control how they want the formatted to behave which enables the user to customize their behavior.

### Context

This issue follows up on #238.

jamesL92:
Closes #453 

This is one possible implementation, with the following example usage:
```sh
configu eval --schema cli.cfgu.json --store foo | configu export --format JSON --eol --formatOptions "pretty=true"
```

This PR should be mutually exclusive to #471 as they both illustrate different ways to solve the same problem.

### DIFF
--- a/packages/cli/src/commands/export.ts
+++ b/packages/cli/src/commands/export.ts
@@ -24,6 +24,7 @@ export const NO_CONFIGS_WARNING_TEXT = 'no configuration was fetched';
 export const CONFIG_EXPORT_RUN_DEFAULT_ERROR_TEXT = 'could not export configurations';
 
 type TemplateContext = { [key: string]: string } | { key: string; value: string }[];
+type FormatOptions = { [key: string]: string };
 
 enum FilterFlag {
   OmitEmpty = 'omit-empty',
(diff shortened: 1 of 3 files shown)


rannn505:
@jamesL92 did you mean `--formatConfig` instead of `--configOptions` (I think the latter will be used as a global cli flag to pass a custom `.configu` file and skip the the auto search of the nearest one)?

jamesL92:
Closing as per discussion in #453

PR #471, opened 2024-10-02
Add --pretty flag to CLI export command

### LINKED ISSUE #453: Merge `CompactJSON` formatter with `JSON` formatter

jamesL92:
Closes #453 

This is one possible implementation, with the following example usage:
```sh
configu eval --schema cli.cfgu.json --store foo | configu export --format JSON --eol --pretty
```

This PR should be mutually exclusive to #470 as they both illustrate different ways to solve the same problem.

### DIFF
--- a/packages/lib/src/formatters/configFormatters.ts
+++ b/packages/lib/src/formatters/configFormatters.ts
@@ -12,6 +12,7 @@ const hasWhitespace = (str: string) => {
 
 type FormatterParameters = {
   json: ExportCommandReturn;
+  pretty?: boolean;
   label: string;
   wrap?: boolean; // * Wraps all values with quotes
 };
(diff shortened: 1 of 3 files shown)


packages/lib/src/formatters/configFormatters.ts
```diff
@@ -78,8 +79,7 @@ const jsonToHelmValues: FormatterFunction = ({ json }) => {
 };
 
 const configFormatters: Record<ConfigFormat, FormatterFunction> = {
-  JSON: ({ json }) => JSON.stringify(json, null, 2),
-  CompactJSON: ({ json }) => JSON.stringify(json),
+  JSON: ({ json }) => JSON.stringify(json),
```
rannn505, line 82: @jamesL92 I would destruct the `pretty` option here, as the formatter should handle it accordingly. If a formatter doesn’t have a "pretty" mode, I think the most intuitive approach would be to simply ignore the value and proceed with the standard formatting.

jamesL92: Good call - this would be a better approach as we expand to more flags.

rannn505 (approved):
LGTM :) thanks @jamesL92

Issue #455, opened 2024-09-23
Configu v1 Repository Structure


rannn505:
- **docs/** (@configu/docs)
- **examples/** (@configu/examples)
- **packages/**
	- **sdk/** (@configu/sdk)
		- Utils - template, expressions, json, json schema
		- Config Interfaces
		- .cfgu - parse, operate
		- Commands - Eval, Upsert, Export
	- **common/** (@configu/common)
		- .configu - locate (cosmiconfig), parse (json-schema), operate (render envs + handle each prop)
		- .cfgu - locate
		- External resource injection - stores, formatters. Cfgu props etc
	- **integrations/**
		- Index.json5
		- stores/
			- @configu/in-memory
			- @configu/aws-parameter-store
		    - …
	    - exporters/
			- @configu/formatters
			- @configu/template
	- **proxy/** (@configu/proxy)
	- **cli/** (@configu/cli)
	- **ui/** (@configu/ui) 
	
```mermaid
flowchart RL
%% Nodes
    A("sdk")
    B("common")
	C("integrations")
	D("cli")
	E("proxy")
	F("ui")

%% Edges
	B --> A
	C --> A
	B --> C

	D --> A
	E --> A
	F --> A

	D --> B
	E --> B
	F --> B

%% Node Styling
    style D color:#FFFFFF, fill:#AA00FF, stroke:#AA00FF
    style E color:#FFFFFF, stroke:#00C853, fill:#00C853
    style F color:#FFFFFF, stroke:#2962FF, fill:#2962FF
```

PR #478, opened 2024-10-02
Allow provision of a default store

### LINKED ISSUE #449: Default Store `.configu`
### Suggestion

- add `default` marker to a store which will be used by default when Store flag is omitted in upper level Interfaces
- Consider change `init` command
- Consider adding a cfgu editor commands

### Motivation

_No response_

### Context

_No response_

jamesL92:
Closes #449 

It'd be nice if we could add some sort of validation when parsing the `.configu` file, to ensure that, among other things, we don't specify two `default` stores. However, since we don't currently have any form of runtime validation of the file, I figured it'd be a bit out of scope for this PR.

### DIFF
--- a/packages/cli/src/base.ts
+++ b/packages/cli/src/base.ts
@@ -21,7 +21,7 @@ type BaseConfig = {
   cli: {
     file?: string; // .configu file
     data: Partial<{
-      stores: Record<string, { type: string; configuration: Record<string, any>; cache?: boolean }>;
+      stores: Record<string, { type: string; configuration: Record<string, any>; cache?: boolean; default?: boolean }>;
       cache?: string;
       schemas: Record<string, string>;
       scripts: Record<string, string>;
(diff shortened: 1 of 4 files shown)


davidmeirlevy:
@jamesL92 thanks for this PR!

@rannn505 can you please review this PR? Thanks!


@jamesL92 By the way, please mention that you take this issue on the issue itself before you start it, so we won't start fixing it ourselves at the same time.

jamesL92:
> @jamesL92 By the way, please mention that you take this issue on the issue itself before you start it, so we won't start fixing it ourselves at the same time.
Yup no probelm, shall do going forward :+1:

rannn505 (approved):
@jamesL92 smooth 🚀

rannn505:
@jamesL92 I'm going to merge this one, but we'll keep the issue open as it will need to be re-implemented in the next generation of the codebase (#455).

PR #501, opened 2024-10-07
typo errors in  hello-world.mdx

Aman1919:
Fixes Typo and Command Line Issues

Corrected file path from /get-started.cfgu.json to ./start.cfgu.json.

### DIFF
--- a/docs/guides/hello-world.mdx
+++ b/docs/guides/hello-world.mdx
@@ -38,25 +38,13 @@ To get started with [Configu Platform](https://app.configu.com/), you'll need to
 ### 4. Upsert Configs to ConfigSet
 
 ```shell
-configu upsert \
-  --store 'configu' --set 'development' --schema './start.cfgu.json' \
-  --config "GREETING=hey" --config "SUBJECT=<YOUR_NAME>" \
-&& configu upsert \
-  --store 'configu' --set 'development/qa' --schema './start.cfgu.json' \
-  -c "GREETING=hola"
-&& configu upsert \
-  --store 'configu' --set 'production' --schema './start.cfgu.json' \
-  -c "GREETING=welcome"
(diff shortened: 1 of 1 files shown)


Aman1919:
I believe there are other files where commands have a trailing slash at the end of the line. Let me know if you're okay with removing the '/' in those cases, or if there's no need for this change, and I’ll proceed accordingly.

docs/guides/hello-world.mdx
```diff
@@ -38,25 +38,13 @@ To get started with [Configu Platform](https://app.configu.com/), you'll need to
...
-configu upsert \
-  --store 'configu' --set 'development' --schema './start.cfgu.json' \
-  --config "GREETING=hey" --config "SUBJECT=<YOUR_NAME>" \
-&& configu upsert \
-  --store 'configu' --set 'development/qa' --schema './start.cfgu.json' \
-  -c "GREETING=hola"
-&& configu upsert \
-  --store 'configu' --set 'production' --schema './start.cfgu.json' \
-  -c "GREETING=welcome"
+configu upsert --store 'configu' --set 'development' --schema './start.cfgu.json' --config "GREETING=hey" --config "SUBJECT=<YOUR_NAME>" && configu upsert --store 'configu' --set 'development/qa' --schema './start.cfgu.json' -c "GREETING=hola" && configu upsert --store 'configu' --set 'production' --schema './start.cfgu.json' -c "GREETING=welcome"
```
jamesL92, line 41: Instead of one-line-ing everything, I think it's more readable to correct the command by adding the missing `\` at the end of line 46 in the original code.

I always find that long commands like the `configu upsert` ones tend to be easier to read across multiple lines rather than a single line. Particularly in IDEs :sweat_smile:

jamesL92 (changes requested):

docs/guides/hello-world.mdx
rannn505: @Aman1919 @jamesL92 is absolutely right—it will be much easier to read as multiline. Thanks for renaming the schema flag! Please keep the multiline string format and fix the missing slash at line 46.

rannn505 (approved):

docs/guides/hello-world.mdx
Aman1919: @rannn505 @jamesL92 thank for your review . i have done the changes

PR #536, opened 2024-10-10
bugfix: disable auto-splitting of aws parameter store

davidmeirlevy:
for some reason, AWS parameter store bundle is splitting into multiple files, which makes an issue if we want to lazy load it from ur public release.
other integrations don't, but I truly recommend adding this property to all integrations for future safety.


RichardAkman (approved):
LGTM

Issue #535, opened 2024-10-10
Verify that every integration build creates one-file only


davidmeirlevy:
### Suggestion

a test, of an after-build step for every integration, to verify that the integration bundle created only one-file.

### Motivation

it seems like some store integrations tend to create a bundle that has more than one file, which leads to runtime errors when we want to lazy-load those integrations on demand.

### Context

when you can "pnpm build:integrations" on the root folder, you can then go to "packages/integrations/dist" and to view all integrations bundles.
the command i mentioned runs all build scripts of all internal integration packages inside the "integrations" library in parallel.
each one creates its own package bundle.

this error might occur if the tsup config is not accurate, or missing external dependencies.

davidmeirlevy:
another simple way to is to add `splitting: false,` to every tsup.config.ts file :)

JayGaba:
hi @davidmeirlevy, can you assign this to me please

PR #547, opened 2024-10-11
Fixed using the simple way suggested

JayGaba:
The fix ensures integration bundle creates only one-file by adding `splitting: false,` to every tsup.config.ts file.

### DIFF
--- a/packages/sdk/tsup.config.ts
+++ b/packages/sdk/tsup.config.ts
@@ -7,7 +7,7 @@ export default defineConfig([
     format: ['esm', 'cjs'],
     dts: true,
     sourcemap: true,
-    splitting: true,
+    splitting: false,
     treeshake: true,
     clean: true,
     keepNames: true,
--- a/packages/cli/tsup.config.ts
+++ b/packages/cli/tsup.config.ts
@@ -11,6 +11,7 @@ export default defineConfig([
     // splitting: true,
     treeshake: true,
     noExternal: [/(.*)/],
+    splitting: false,
     // https://github.com/egoist/tsup/issues/939
     // outExtension: () => {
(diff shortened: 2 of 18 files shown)


JayGaba:
#Closes #535

JayGaba:
@davidmeirlevy please review and merge, thanks!

packages/sdk/tsup.config.ts
```diff
@@ -7,7 +7,7 @@ export default defineConfig([
     format: ['esm', 'cjs'],
     dts: true,
     sourcemap: true,
-    splitting: true,
+    splitting: false,
```
davidmeirlevy, line 10: The sdk is not part of the integrations. It can split the bundle if it wants to.

davidmeirlevy (changes requested):
Thanks for your PR!
Just one change and we can merge.

packages/sdk/tsup.config.ts
JayGaba: Done! Please review and let me know if any further changes are required.

davidmeirlevy (approved):

davidmeirlevy:
It seems like the lock file should be updated accordingly.
Please run "pnpm install" to create a new lock file, and commit it, so the CI could pass.

JayGaba:
> It seems like the lock file should be updated accordingly. Please run "pnpm install" to create a new lock file, and commit it, so the CI could pass.

apologies @davidmeirlevy , i think it should pass now!

davidmeirlevy:
@JayGaba merged! thanks for your contribution 👏

JayGaba:
> @JayGaba merged! thanks for your contribution 👏

Thank you for making my first contribution so smooth! 🚀

PR #548, opened 2024-10-13
Support glob patterns in .configu schema definitions

### LINKED ISSUE #440: Enable glob-based paths for `ConfigSchema`s
### Suggestion

Enable glob-based paths for `ConfigSchema`s in the `.configu` file and with the `--schema` flag on the CLI. When a glob path is used, the interface will expand it, and if multiple schemas are found, it will combine them. If a key is in multiple schemas, the "deepest" one takes precedence.

### Motivation

Simplifies managing multiple schemas by allowing pattern-based file selection. Useful for projects with hierarchical or scattered configurations. Ensures specific configurations override general ones.

### Context

Currently, users must specify each schema file individually, which can be cumbersome and error-prone, especially in large projects with many configuration files. By allowing glob-based paths, we streamline the process, making it easier to manage and apply multiple schemas. This is particularly useful in environments where schemas are organized in a directory structure that reflects their scope and specificity.

For instance, a user might have a directory structure where common schemas are in a top-level directory, and more specific schemas are nested deeper. By using a glob pattern, such as schemas/**/*.yaml, the interface will automatically include all relevant schemas. If a key is declared in multiple schemas, the rule that the "deepest" one wins ensures that the most specific configuration is applied.

jamesL92:
Closes #440

### DIFF
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -52,6 +52,7 @@
     "@configu-integrations/toml": "workspace:*",
     "@configu-integrations/configu": "workspace:*",
     "@configu-integrations/json-file": "workspace:*",
+    "fast-glob": "^3.3.2",
     "find-up": "^7.0.0",
     "js-yaml": "^4.1.0",
     "lodash": "^4.17.21",
--- a/packages/common/src/utils.ts
+++ b/packages/common/src/utils.ts
@@ -65,8 +65,3 @@ export const parseYAML = (filePath: string, fileContent: string): any => {
     throw error;
   }
 };
-
-export const mergeSchemas = (...schemas: ConfigSchema[]): ConfigSchema => {
-  // Later schemas take precedence in case of key duplication.
-  return new ConfigSchema(_.merge({}, ...schemas.map((schema) => schema.keys)));
-};
--- a/packages/common/src/utils.ts
+++ b/packages/common/src/utils.ts
@@ -3,6 +3,7 @@ import path from 'pathe';
 import _ from 'lodash';
 import parseJson from 'parse-json';
 import yaml from 'js-yaml';
+import { ConfigSchema } from '@configu/sdk';
 
 export const { NODE_ENV } = process.env;
 export const isDev = NODE_ENV === 'development';
@@ -63,4 +64,9 @@ export const parseYAML = (filePath: string, fileContent: string): any => {
     error.message = `YAML Error in ${filePath}:\n${error.message}`;
     throw error;
   }
+};
+
+export const mergeSchemas = (...schemas: ConfigSchema[]): ConfigSchema => {
+  // Later schemas take precedence in case of key duplication.
+  return new ConfigSchema(_.merge({}, ...schemas.map((schema) => schema.keys)));
 };
--- a/packages/sdk/src/core/ConfigSchema.ts
+++ b/packages/sdk/src/core/ConfigSchema.ts
@@ -26,6 +26,10 @@ export class ConfigSchema {
         // }
       });
   }
+
+  static merge(...schemas: ConfigSchema[]): ConfigSchema {
+    return new ConfigSchema(_.merge({}, ...schemas.map((schema) => schema.keys)));
+  }
 }
 
 export const ConfigSchemaKeysSchema: JsonSchemaType<ConfigSchemaKeys> = {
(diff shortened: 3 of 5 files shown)


packages/sdk/src/core/ConfigSchema.ts
```diff
@@ -26,6 +26,10 @@ export class ConfigSchema {
         // }
       });
   }
+
+  static merge(...schemas: ConfigSchema[]): ConfigSchema {
+    return new ConfigSchema(_.merge({}, ...schemas.map((schema) => schema.keys)));
+  }
```
RichardAkman, line 32: I suggest we keep this logic contained within the common package for the glob patterns feature. The internal SDK should be as lean as possible to keep the architecture of the project as straightforward as possible.

RichardAkman (changes requested):
Really solid implementation ⭐!

I did notice that certain patterns can cause the fast-glob library to hang, though these are due to incorrect formatting. For the purposes of this feature, the current implementation works well. However, we’ll need to ensure users are aware of how to properly use it to prevent the interface from hanging, which could lead to a frustrating UX.

It might be worth exploring whether we can customize the library's behavior to handle such cases more gracefully. We can consider extending this logic based on usage feedback.

Please review this comment, and once addressed, we can get this merged 🥇.

packages/sdk/src/core/ConfigSchema.ts
jamesL92: Makes sense :) I'll look to make this change sometime this evening.

pnpm-lock.yaml
jamesL92, line 1: It appears that `pnpm-lock.yaml` is currently out of date on `main`, which I started seeing after rebasing. Happy to handle the cleanup in a separate PR if you'd like, however as I'm adding the `fast-glob` package to `common` here the lock file would need updating regardless.

packages/common/src/utils.ts
```diff
@@ -64,3 +65,8 @@ export const parseYAML = (filePath: string, fileContent: string): any => {
     throw error;
   }
 };
+
+export const mergeSchemas = (...schemas: ConfigSchema[]): ConfigSchema => {
+  // Later schemas take precedence in case of key duplication.
```
jamesL92, line 70: It kinda bugs me that this is implicitly the case, but figured this might be OK for a first iteration.

Honestly I think I'd prefer to have the behaviour such that two conflicting configSchemas (i.e. at the same depth, with the same key) explicitly throws an error - I feel like it'd be less likely to lead to potentially confusing bugs for the user. Would love to get some feedback on the team's thoughts around this.

packages/common/package.json
```diff
@@ -52,6 +52,7 @@
     "@configu-integrations/toml": "workspace:*",
     "@configu-integrations/configu": "workspace:*",
     "@configu-integrations/json-file": "workspace:*",
+    "fast-glob": "^3.3.2",
```
rannn505, line 55: @jamesL92 can you please attach an npm trends or similar comparison between glob libs 🙏🏽  why did you pick `fast-glob`?

packages/common/src/utils.ts
rannn505: @jamesL92 I planing on opening a new feature for implementing an `extend` keyword for the `.cfgu` file which practicaly need this merge method, with that in mind do you think it is maybe better to put it on the CfguFile Class as a static method?

cc @configu/dev

rannn505 (approved):
@jamesL92 This is truly marvelous work! I can't thank you enough! 🙌

rannn505 (approved):
@jamesL92 This is truly marvelous work! I can't thank you enough! 🙌

packages/common/src/utils.ts
jamesL92: can do, yup :)

packages/common/package.json
jamesL92: Picked `fast-glob` as it _was_ already being used in the repository before (see [here](https://github.com/configu/configu/blob/8eca84c9a30e1e15599251b19440df7e2553d0df/packages/cli/package.json#L111), specifically was used quite a bit for the old [find](https://github.com/configu/configu/blob/8eca84c9a30e1e15599251b19440df7e2553d0df/packages/cli/src/commands/find.ts) command prior to migration).

Can see it's no longer being used by the CLI since the migration to `clipanion`, although I guess that's mainly because the `find` command hasn't been ported yet. As such, I'm happy to take a look into a comparison between different libraries if you'd like :+1:

RichardAkman (approved):
LGTM, great work!

PR #549, opened 2024-10-13
Add support for default store in new repo structure

### LINKED ISSUE #449: Default Store `.configu`

jamesL92:
Closes #449, using the new repository structure.


packages/common/src/ConfiguFile.ts
```diff
@@ -120,16 +125,21 @@ export class ConfiguFile {
     return ConfiguFile.load(path);
   }
 
-  getStoreInstance(name: string) {
-    const storeConfig = this.contents.stores?.[name];
+  getDefaultStoreName() {
+    const defaultStoreName = _.findKey(this.contents.stores, (store) => store.default);
```
rannn505, line 129: @jamesL92 what do you think about a case when a user only have 1 store defined in its .configu file, should we treat it as a "default" and change the behiour of the "defailt-only export" when executing configu export only with --schema to --schema --defaults or somthing similar?

cc @configu/dev

jamesL92: Ooh I like the idea of auto-defaulting a store if there's only one present, fantastic idea!

What do you mean by "default-only export"? Not quite sure I understand.

packages/common/src/ConfiguFile.ts
```diff
@@ -5,13 +5,16 @@ import { findUp } from 'find-up';
 import { ConfigSchema, ConfigStore, Expression, JsonSchema, JsonSchemaType } from '@configu/sdk';
 import FastGlob from 'fast-glob';
 import _ from 'lodash';
-import { readFile, parseJSON, parseYAML, mergeSchemas } from './utils';
```
jamesL92, line 8: Cleanup of a bug I introduced on a previous PR (surfaced as a merge conflict when rebasing)

packages/common/src/ConfiguFile.ts
rannn505: The current state is that if a store is not supplied via the cli eval command we are using the noop store which basically ends up evaluating all keys of the used .cfgu file to their default or empty values

packages/common/src/ConfiguFile.ts
rannn505: Good catch

packages/common/src/ConfiguFile.ts
jamesL92: Ah right I got you - yeah I've added in the logic now to treat a single store in the the `.configu` file as the default store - so the "default-only export" will now have the following behaviour:
- If `stores` not provided in `.configu`, or there are no stores configured, we will use the `noop` store.
- If only 1 store is configured, that store will be used (as a single store is auto-treated as the default).
- If multiple stores are configured
  - If at least 1 marked as `default`, then the first store marked as `default` will be used.
  - If no stores marked as `default`, then the `noop` store will be used.

Happy to change this behaviour if something different is desired. :+1:

rannn505 (approved):
Smooth! 🥇

PR #550, opened 2024-10-13
initial load of integrations that are included inside the cli and proxy

davidmeirlevy:


packages/common/package.json
```diff
@@ -45,6 +45,11 @@
   },
   "dependencies": {
     "@configu/sdk": "workspace:*",
+    "@configu-integrations/validator": "workspace:*",
+    "@configu-integrations/comptact-json": "workspace:*",
```
RichardAkman, line 49: Should be `compact` instead of `comptact`

packages/common/package.json
```diff
@@ -45,6 +45,11 @@
   },
   "dependencies": {
```
RichardAkman, line 46: We're missing `configu` store here that should also be inbuilt

packages/common/package.json
davidmeirlevy: fixed.

packages/common/package.json
davidmeirlevy: fixed/

RichardAkman (approved):
LGTM

Issue #503, opened 2024-10-07
Stablize `@configu/cli`


rannn505:
### Suggestion

Refactor the `@configu/cli` framework from `oclif` to `clipanion`. Clipanion is a newer, thinner alternative (with only 1 dependency compared to `oclif`'s 17) and is used by high-profile CLIs like Corepack, Yarn, and PNPM. Clipanion can also be bundled into a single file, allowing us to later use `node-sea` to create a true binary for the CLI.

### Motivation

Moving to `clipanion` reduces the dependency footprint and simplifies the CLI architecture. It also opens the door for better packaging and distribution options, like bundling into a single file for easier deployment. This will improve performance, maintainability, and flexibility.

### Context

- next framework: https://mael.dev/clipanion/
- old framework: https://oclif.io/
- other potential alternative: https://bloomberg.github.io/stricli/

- https://nodejs.org/api/single-executable-applications.html
- https://github.com/nodejs/single-executable
- https://github.com/nodejs/single-executable/blob/main/docs/production-nodejs-clis.md
- https://github.com/liudonghua123/node-sea/
- https://dev.to/chad_r_stewart/compile-a-single-executable-from-your-node-app-with-nodejs-20-and-esbuild-210j
- https://github.com/arcanis/node-sea
- https://github.com/bryopsida/node-sea-action/

jamesL92:
@rannn505 please assign to me, I'd be more than happy to take a look at this :)

RichardAkman:
List of issues/PRs that are a part of this issue:
- #545
- #565
- #570
- #568
- #609
- #602
- #606 OR #581

PR #568, opened 2024-10-15
Complete Configu CLI - Export command

RichardAkman:
Partially completes #503 by implementing export command

Usage examples:

Filter flag:
```bash
tsx packages/cli/src/run.ts eval --store store1 --set '' --schema ./tmp/schemas2/start.cfgu.yaml | tsx packages/cli/src/run.ts export --filter 'empty($.value)' --filter 'contains($.cfgu.label,"a")'
```

Template flag:
```bash
tsx packages/cli/src/run.ts eval --store store1 --set '' --schema ./tmp/schemas2/start.cfgu.yaml | tsx packages/cli/src/run.ts export --template ./tmp/template.env
```

Format flag:
```
tsx packages/cli/src/run.ts eval --store store1 --set '' --schema ./tmp/schemas2/start.cfgu.yaml | tsx packages/cli/src/run.ts export --format 'Dotenv({json:{{pipe}},wrap:true})'
```

Schema used in example:
```yaml
keys:
  GREETING:
    default: hello
  SUBJECT:
    default: world
    label:
      - 'a'
  MESSAGE:
    description: Generates a full greeting message
```

Template used example:
```Dotenv
SUBJECT={{SUBJECT}}
GREETING={{toUpper(GREETING)}}
```


packages/cli/src/commands/export.ts
```diff
@@ -19,20 +71,217 @@ export class ExportCommand extends BaseCommand {
...
+      return;
+    }
 
+    const label = this.label ?? `configs-${Date.now()}`;
+    const filter = this.filterFromFlags();
+    const keys = this.keysMutations();
+    const pipe = keys
+      ? _.mapValues(previousEvalCommandOutput, (config, key) => ({ ...config, key: keys(key) }))
+      : previousEvalCommandOutput;
+    // TODO: how do we implement the format flag? we somehow have to connect the format flag with the formatter from the registry and pass it via the 'reduce' prop of ExportCommand
```
RichardAkman, line 280: @davidmeirlevy, I need your input here. When providing the `format` flag, we need to fetch the formatter from the registry or download it remotely somehow but it currently doesn't expose expressions or any other logic to do this. How do you think we should implement this logic?

packages/cli/src/commands/export.ts
```diff
@@ -19,20 +71,217 @@ export class ExportCommand extends BaseCommand {
...
+    const exportCommand = new BaseExportCommand({ pipe, filter });
+    const { result } = await exportCommand.run();
+    // TODO: the previous implementation expected the result to be an object but now export command returns a string that needs to be parsed.
```
RichardAkman, line 283: @davidmeirlevy, ExportCommand now returns a plain string that's reduced based on the `reduce` prop. The issue is that `exportConfigs` wants the result as a plain object for the many other flags that need to use the result such as the `template` flag. This means we will need to parse the export command result which means we now need a parser for each format which feels very redundant. How do you feel we should go about solving this? maybe we should reconsider the ExportCommand signature

packages/cli/src/commands/export.ts
```diff
@@ -19,20 +71,217 @@ export class ExportCommand extends BaseCommand {
...
+
+      const isKeyOmitted = omittedKeys.includes(key);
+      const isLabelOmitted = _.intersection(omittedLabels, cfgu.label ?? []).length > 0;
+      const omitFilter = !isKeyOmitted && !isLabelOmitted;
+
+      return hiddenFilter && emptyFilter && pickFilter && omitFilter;
+    };
+  }
+
+  async exportConfigs(result: ExportCommandOutput, label: string) {
```
RichardAkman, line 222: This function needs to get the result of export command after parsing.

packages/cli/src/commands/export.ts
```diff
@@ -19,20 +71,217 @@ export class ExportCommand extends BaseCommand {
     // ],
   });
 
-  // name = Option.String(''
+  format = Option.String('--format', {
+    description: `Format exported \`Configs\` to common configuration formats. Redirect the output to file, if needed`,
+    // TODO: get this list from somewhere? or maybe let this fail during runtime if not found in registry?
```
RichardAkman, line 76: @davidmeirlevy, how do you feel we should handle this?

davidmeirlevy (approved):

Issue #572, opened 2024-10-16
String key value returning as empty string instead of actual value


LongBeachHXC:
### Current Behavior

Seems like there could be a bug in the `ini-file` formatter. When using the ini file store I get an empty value when trying to export the test key var. When using the json file store, I get the correct value when exporting.

### Expected Behavior

I would expect configu to return "false" just like when using the json file store as a backend.

### To Reproduce

```
configu --version
@configu/cli/0.17.1 wsl-x64 node-v20.10.0

configu upsert --store localJson --set test --schema test.cfgu.yaml --config "TEST_KEY=false"
configu upsert --store localIni --set test --schema test.cfgu.yaml --config "TEST_KEY=false"

configu eval --store localJson --set test --schema test.cfgu.yaml | configu export --format YAML
TEST_KEY: 'false'

configu eval --store localIni --set test --schema test.cfgu.yaml | configu export --format YAML
TEST_KEY: ''
```

```
# config.cfgu.yaml
---
TEST_KEY:
  description: This is just an example
  type: String
...
```
```
# .configu
{
  "stores": {
    "localJson": {
      "type": "json-file",
      "configuration": {
        "path": "config-store.json"
      }
    },
    "localIni": {
      "type": "ini-file",
      "configuration": {
        "path": "config-store.ini"
      }
    }
  }
}
```

jamesL92:
Hi @LongBeachHXC. Thanks for reporting this bug. 
The Configu CLI is currently undergoing a full rewrite, so I doubt this version of the CLI will be retroactively fixed. The `ini` store has yet to be migrated to the new CLI implementation, however there is an open issue to implement it (see #516). I'll test this bug report against the new implementation of the store to see if the bug persists, and will get back to you then.

jamesL92:
I've debugged the issue, turns out the root cause is that the library being used to parse the `ini` file will treat boolean values as boolean, rather than strings. This in turn makes Configu's logic skip over the value.

Probably the most robust solution is when writing to an ini file, we want to make sure we escape all the config values. So after running
```bash
configu upsert --store localIni --set test --schema test.cfgu.yaml --config "TEST_KEY=false"
```
we should see the following content:
```
# config-store.ini
[test]
TEST_KEY="false"
```
instead of:
```
# config-store.ini
[test]
TEST_KEY=false
```

RichardAkman:
[...]

PR #573, opened 2024-10-17
Migrate IniFileConfigStore to @configu/integrations

### LINKED ISSUE #516: Migrate `IniFileConfigStore` to `@configu/integrations`
### Suggestion

Migrate the `IniFileConfigStore` from its current setup to the new `@configu/integrations` module as part of the major restructuring. The goal is to move all third-party integrations, including stores, expressions, and libraries, to `@configu/integrations`.

### Motivation

This migration is part of a larger effort to better organize and modularize the product. Moving the `IniFileConfigStore` will align with the new structure, making the system cleaner and more maintainable.

### Context

You can refer to the already migrated stores for guidance:  
[Migrated stores in integrations](https://github.com/configu/configu/tree/main/packages/integrations/stores).

Base stores implementations from `@configu/ts` can be found here:  
[Base stores in TS](https://github.com/configu/configu/tree/52cee9c41fb03addc4c0983028e37df42945f5b7/packages/ts/src/stores).

The old node-based stores were previously in `@configu/node`:  
[Node stores](https://github.com/configu/configu/tree/52cee9c41fb03addc4c0983028e37df42945f5b7/packages/node/src/stores).

The old browser-based stores were previously in `@configu/browser`:  
[Browser stores](https://github.com/configu/configu/tree/52cee9c41fb03addc4c0983028e37df42945f5b7/packages/browser/src/stores).

i-sanjay-cs:
Closes #516


packages/integrations/stores/ini-file/package.json
```diff
@@ -0,0 +1,35 @@
+{
+    "name": "@configu-integrations/ini-file",
+    "version": "0.0.0",
+    "license": "Apache-2.0",
+    "author": "@configu/dev",
+    "files": [],
```
jamesL92, line 6: Just did some testing, you'll need to change this to:
```suggestion
    "main": "src/ini-file.ts",
```
in order for this to successfully register as an integration

packages/integrations/stores/ini-file/src/ini-file.ts
```diff
@@ -0,0 +1,44 @@
+import _ from 'lodash';
```
jamesL92, line 1: `lodash` is missing from this integration's `package.json`. Could you please add it? N.B. it's pinned elsewhere to `^4.17.21', so I'd keep to that version.

jamesL92 (changes requested):
Just did some testing, looks like the bug referenced in #572 is still present in this implementation.

@rannn505 would you like the bug fixed as part of this PR, or as a separate PR once this is merged in?

packages/integrations/stores/ini-file/src/ini-file.ts
```diff
@@ -0,0 +1,44 @@
...
+import * as ini from 'ini';
+import { type Config } from '@configu/sdk';
+import { FileConfigStore } from '@configu/integrations/src/utils/File';
+
+export type IniFileConfigStoreConfiguration = { path: string };
+
+export class IniFileConfigStore extends FileConfigStore {
+    constructor({ path }: IniFileConfigStoreConfiguration) {
+        const initialFileState = '';
+        super('ini-file', { path, initialFileState });
```
jamesL92, line 11: The implementation of `FileConfigStore` has changed, there's no need to pass in the `type` parameter any more. This can be fixed with:
```suggestion
        super({ path, initialFileState });
```

RichardAkman:
@jamesL92 thanks for the quality review! Regarding #572, I suggest we handle it in a separate PR as it's not relevant to the migration of the config store, it should be handled after the migration is complete to keep the scope of this PR lean.

i-sanjay-cs:
@RichardAkman  do i need to change anything in this ?

RichardAkman:
@i-sanjay-cs, all of the change requests that @jamesL92 made are relevant besides the one referred to in #572. Please apply them before we proceed with the review 🙏🏻

packages/integrations/stores/ini-file/package.json
jamesL92: Actually most of the other integrations have `"files": [],`, so maybe this isn't actually needed. Not sure why it was failing for me during testing but I'd say feel free to disregard this suggestion for the time being.

If it fails to register we can take a look as part of a separate issue, but for the time being I'd say consistency is better.

RichardAkman:
Hi @i-sanjay-cs, I noticed you've unassigned yourself from this PR, and there haven’t been any updates recently. Could you let me know if you plan to continue working on it? If not, I’ll close the PR to make the issue available for other contributors. Thank you!

i-sanjay-cs:
Hey @RichardAkman I am little bit busy with the current work thats why I just unassigned so any other can work on this issue. So you can close it.

PR #580, opened 2024-10-20
testbed for configu integrations

davidmeirlevy:


packages/integrations/src/test-utils/index.ts
```diff
@@ -0,0 +1,65 @@
+import test, { describe, beforeEach } from 'node:test';
+import * as assert from 'node:assert';
+import { Test, TestBed } from './test-bed';
+
+describe('example', () => {
+  let module: TestBed;
+  const mockStore = Test.createMockStore();
```
jamesL92, line 7: I'd recommend instantiating this in `beforeEach` - that way the call count will be reset on each test, rather than relying on the tests being run in a specific order.

davidmeirlevy: Yes of course.
I just want to demonstrate the functions that I want to implement as part as the test utils.

Are those testing abilities looks legit?

RichardAkman:
In preparation for v1, Ran and I have agreed to tidy up all open PRs. Since the testbed still requires additional work, we've decided to archive it for now and revisit it when it becomes a priority. I'll close this PR accordingly.

Issue #528, opened 2024-10-09
Implement `configu generate` Command in `@configu/cli`


rannn505:
### Suggestion

Implement the `configu generate` command with `--dir` and `--name` flags. This command will create an integration module in the specified directory, ready to be injected into the system.

### Motivation

Simplifies the creation of integration modules by automating the setup process. Developers can quickly generate the necessary files and structure with a single command, improving productivity and reducing manual setup effort.

### Context

The `configu generate` command will take a directory (`--dir`) and module name (`--name`) as inputs, creating a pre-configured integration module that is ready for injection into the project.

lyle45:
I would like to do it

davidmeirlevy:
@lyle45 you got it :)

davidmeirlevy:
@lyle45 are you still on it?
because we want to deliver this feature soon.

lyle45:
@davidmeirlevy hi David. Sorry was preoccupied. Will do it today hopefully

davidmeirlevy:
if you need any help with that, don't hesitate to reach me on our discord channel.

davidmeirlevy:
@lyle45 I see that it takes too much, and this is a very core feature of configu that we must deliver very soon.
So I need to re-assign it.

jamesL92:
Im happy to pick this up :)

davidmeirlevy:
@jamesL92 you can checkout from my branch..

lyle45:
@davidmeirlevy yeah my apologies. Couldn't get to it. Sorry

Issue #465, opened 2024-09-29
Configu Hacktoberfest 2024


pelegpor:
### Suggestion

Configu is thrilled to be part of Hacktoberfest 2024 ❤️

RS-labhub:
Will you be opening new issues or do we have to find it ourselves?

RichardAkman:
Hi @RS-labhub, we’ll be adding new issues very soon, so stay tuned! In the meantime, feel free to browse the existing issues or create new ones as you see fit!

Dxuian:
> Hi @RS-labhub, we’ll be adding new issues very soon, so stay tuned! In the meantime, feel free to browse the existing issues or create new ones as you see fit!

will they be added under hacktoberfest tag ?

RS-labhub:
> > Hi @RS-labhub, we’ll be adding new issues very soon, so stay tuned! In the meantime, feel free to browse the existing issues or create new ones as you see fit!
> 
> will they be added under hacktoberfest tag ?

Yes!

gagandeepp:
@RichardAkman how to claim configu swag as my MR has been merged?

RichardAkman:
@gagandeepp we will provide all necessary details after Hacktoberfest ends, stay tuned!

abhirajadhikary06:
Thanks for the update 👍

RichardAkman:
> Just a bit curious, here it's written we will get a tee and a beer glass for 3+ merged PRs, but in the mentioned issue, it's just written 3 to get both of them... Which one is correct?
> 
> mentioning: @RichardAkman and @pelegpor

@RS-labhub to clarify, 3 is enough to get both the rewards.

RS-labhub:
Ehehe.. I'll be getting both now 🥂🌊

gagandeepp:
@RichardAkman Any update on claiming the swag

RichardAkman:
Hi @gagandeepp, thanks for reaching out! We’re actively working to finalize this year’s Hacktoberfest efforts. As noted in the main post, rewards will be shipped within 30 to 180 days following Hacktoberfest, meaning after November 1, 2024. We’ll provide more details and instructions for claiming your rewards soon. Thanks for your patience and contributions!

i-sanjay-cs:
Hello @RichardAkman any updates regarding the hacktoberfest rewards.

pelegpor:
Hi @i-sanjay-cs

[...]

PR #606, opened 2024-10-21
Initial implementation of the generate command

jamesL92:
Partially solves #528.

As a first pass over this I figured it'd make sense to have the command manually copy out the contents rather than download the boilerplate from a remote location (mainly so we can actually test it out).

A second pass can then either enhance the this functionality, or modify it to download the project boilerplate from a remote location. I think it's worth further discussion/consideration as to which path to take before we commit to one or the other.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -50,7 +50,8 @@
     "openid-client": "5.7.0",
     "std-env": "^3.7.0",
     "tty-table": "4.2.3",
-    "typanion": "3.14.0"
+    "typanion": "3.14.0",
+    "tiged": "2.12.7"
   },
   "bundleDependencies": [],
   "devDependencies": {}
--- a/packages/cli/src/commands/generate.ts
+++ b/packages/cli/src/commands/generate.ts
@@ -1,0 +1,69 @@
+import { Command, Option } from 'clipanion';
+import { cp, readFile, writeFile, rename } from 'node:fs/promises';
+import { dirname, join } from 'node:path';
+import { spawnSync } from 'node:child_process';
+import { pascalCase } from 'change-case';
+import { BaseCommand } from './base';
+
+export class GenerateCommand extends BaseCommand {
+  static override paths = [['generate'], ['gen']];
+
+  static override usage = Command.Usage({
+    // category: `My category`,
+    description: `Generate a boilerplated project for a custom \`ConfigStore\``,
+    // details: `
+    //   A longer description of the command with some \`markdown code\`.
+
+    //   Multiple paragraphs are allowed. Clipanion will take care of both reindenting the content and wrapping the paragraphs as needed.
+    // `,
+    // examples: [
+    //   [`A basic example`, `$0 my-command`],
+    //   [`A second example`, `$0 my-command --with-parameter`],
+    // ],
+  });
+
+  directory = Option.String('--dir,--di', {
+    description: `Location to generate the \`ConfigStore\` project`,
+    required: true,
+  });
+
+  name = Option.String('--name,--na', {
+    description: `Name of the \`ConfigStore\``,
+    required: true,
+  });
+
+  async execute() {
+    await this.init();
+
+    console.log(`Copying from ${join(dirname(import.meta.dirname), 'integration_template')} to ${this.directory}`);
+    console.log('Generating project...');
+    await this.generateBoilerplate();
+    await this.overrideTemplates();
+    await this.installDependencies();
+  }
+
+  private async generateBoilerplate() {
+    console.log(`Copying project boilerplate...\r`);
+    await cp(join(dirname(import.meta.dirname), 'integration_template'), this.directory, { recursive: true });
+    console.log(`Copying project boilerplate...done`);
+  }
+
+  private async overrideTemplates() {
+    console.log(`Overriding to match your selected name...\r`);
+    rename(join(this.directory, 'src', '{{name}}.ts'), join(this.directory, 'src', `${this.name}.ts`));
+    await this.rewriteTemplateFile(join(this.directory, 'src', `${this.name}.ts`));
+    await this.rewriteTemplateFile(join(this.directory, `package.json`));
+    console.log(`Overriding to match your selected name...done`);
+  }
+
+  private async rewriteTemplateFile(path: string) {
+    const content = await readFile(path, 'utf-8');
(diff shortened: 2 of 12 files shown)


jamesL92:
Going to start a discussion here, as the context of this PR might be useful for framing the discussion.

There are two approaches we could take to writing out the project boilerplate, each with their own tradeoffs:
1 - Have the `generate` command write out all the necessary content itself.
2 - Have the `generate` command download the relevant content from a remote location (e.g. directly from GitHub).

On one hand, option 2 makes it easier for us to keep boilerplate up to date, as (for example) the user would not need to update their CLI to gain access to updated boilerplate.

On the other hand, option 1 keeps the dependency tree slimmer, as we wouldn't need a new dependency ([tiged](https://github.com/tiged/tiged) was suggested as one such candidate) which in theory should reduce the bundle size. Additionally, it allows us to version the boilerplate alongside individual CLI versions - so we wouldn't need to worry about compatibility issues if breaking changes are introduced.

jamesL92:
Also @davidmeirlevy apologies, wasn't sure how to pull your branch into my fork to build on top of it, so had to start my own branch from scratch.

packages/cli/src/commands/generate.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  name = Option.String('--name,--na', {
+    description: `Name of the \`ConfigStore\``,
+    required: true,
+  });
+
+  async execute() {
+    await this.init();
+
+    console.log(`Copying from ${join(dirname(import.meta.dirname), 'integration_template')} to ${this.directory}`);
```
jamesL92, line 38: Happy to change any of the `console.log` contents here, but figured it'd be nice to have something there to give the user some sense of progress.

packages/cli/src/commands/generate.ts
```diff
@@ -0,0 +1,69 @@
...
+  private async rewriteTemplateFile(path: string) {
+    const content = await readFile(path, 'utf-8');
+    const newContent = content.replaceAll('{{name}}', this.name).replaceAll('{{name:pascal}}', pascalCase(this.name));
+    await writeFile(path, newContent, 'utf-8');
+  }
+
+  private async installDependencies() {
+    console.log(`Installing dependencies...\r`);
+    spawnSync('pnpm', ['install'], { cwd: this.directory, stdio: 'inherit' });
+  }
```
jamesL92, line 68: I was torn about whether or not we want to do this - on one hand it's one less thing for the user to do, on the other hand we might want to allow the user to choose their own package manager of choice (e.g. `npm` or `yarn`)

packages/cli/src/commands/generate.ts
```diff
@@ -0,0 +1,69 @@
...
+    console.log(`Copying from ${join(dirname(import.meta.dirname), 'integration_template')} to ${this.directory}`);
+    console.log('Generating project...');
+    await this.generateBoilerplate();
+    await this.overrideTemplates();
+    await this.installDependencies();
+  }
+
+  private async generateBoilerplate() {
+    console.log(`Copying project boilerplate...\r`);
+    await cp(join(dirname(import.meta.dirname), 'integration_template'), this.directory, { recursive: true });
```
jamesL92, line 47: If we go down the `write-from-local` route rather than downloading the boilerplate, we'll probably want to bring all the contents into strings in JS files rather than `cp`-ing them - since we want the command line tool to bundle to a single executable.

davidmeirlevy:
@jamesL92 that solution won't work.

There's a reason why we specified using "degit" with this feature:
The CLI is a compiled executable tool.

Which means: we bundle all the code, including the dependencies in one-file, including node.js executable itself.
So, libraries and folders cannot be included inside of it.

We do it because most of the users are not js developers, and they don't have node.js on their machine.

There are also other secondary reasons:
1. We want the CLI to executable to be as lean as possible, so we don't include anything that can make the CLI heavier. The generate command is something that most devs won't use. Most users will probably use integrations that made by others (common to any OSS).
2. We would like to have the ability to improve the integration boilerplate without forcing our users to manually upgrade the CLI.
3. Using "degit" to import remote boilerplate is a very common practice in many js OSS projects.

packages/cli/src/commands/generate.ts
rannn505: we can use corepack or something but i wouldnt run anything maybe we should print the "next instructions" in a box as a list after directory is generated.

packages/cli/src/commands/generate.ts
rannn505: we can potentially tight between the boilerplate dir / repo to the cli deployment in a way that they will be bundled as artifacts with the tsup of the cli / or use digit as a dev dependency and trigger it as part of the cli release process to download and bundled them.

packages/cli/src/commands/generate.ts
rannn505: use the this.stdio (based on the `consola` dependency)

packages/cli/package.json
```diff
@@ -39,13 +39,15 @@
...
+    "change-case": "4.1.2",
     "clipanion": "^4.0.0-rc.4",
     "consola": "^3.2.3",
     "inquirer": "12.0.0",
     "lodash": "^4.17.21",
     "open": "10.1.0",
     "openid-client": "5.7.0",
-    "std-env": "^3.7.0"
+    "std-env": "^3.7.0",
+    "tiged": "2.12.7"
```
jamesL92, line 50: Tiged is a fork of degit, which is still actively maintained (last update [2 months ago](https://github.com/tiged/tiged) rather than [3 years ago](https://github.com/Rich-Harris/degit))

davidmeirlevy: 👍

davidmeirlevy (approved):
lgtm

davidmeirlevy:
@jamesL92 looks good!

please rebase and fix the lock conflicts, and we'll merge it.

jamesL92:
@davidmeirlevy done, this should be good to merge now 👍

pelegpor:
Hi @jamesL92 ,

Here are the details for claiming your Hacktoberfest prizes! 🎉 Check out [#465  (comment)](https://github.com/configu/configu/issues/465#issuecomment-2498550259) for more information or simply fill out the form here: [Claim Your Swag](https://share-eu1.hsforms.com/2SVc5VKY3Thi04Bvjh17WKw2e62o0).  

Remember, you only need to fill out the form once for all your PRs. 😊

PR #635, opened 2024-10-28
Migrate `Configu init` command to `@configu/cli`

### LINKED ISSUE #621: Migrate `Configu Init` command to `@configu/cli`
### Suggestion

Return the previous `Configu Init` command

### Motivation

Until the `Configu Init` command is reproductized by @rannn505, we should have the previous init command for the CLI feature completeness.

### Context

Requested by @pelegpor.

RichardAkman:
Closes #621

Notes:
- All references and related logic to the old `type` property was removed from the migrated command
- Migrated some logic from the old `@configu/lib` that are necessary for the command
- The schema generated is adapted to the new version of `ConfigSchema` that's expected by `@configu/common`. The template is commented out for now until the property is returned by `@configu/sdk`
- Fixes an issue in `base.ts` where the `Option` parameter wasn't imported

### DIFF
--- a/packages/cli/src/helpers/configExtractor.ts
+++ b/packages/cli/src/helpers/configExtractor.ts
@@ -1,0 +1,100 @@
+import Dotenv from 'dotenv';
+import _ from 'lodash';
+import { Cfgu, Config } from '@configu/sdk';
+
+type ConfigFormat =
+  | 'JSON'
+  | 'CompactJSON'
+  | 'YAML'
+  | 'Dotenv'
+  | 'KubernetesConfigMap'
+  | 'HelmValues'
+  | 'TerraformTfvars'
+  | 'TOML';
+
+const CONFIG_FORMAT_EXTENSION: Record<ConfigFormat, string> = {
+  JSON: 'json',
+  CompactJSON: 'json',
+  YAML: 'yaml',
+  Dotenv: 'env',
+  KubernetesConfigMap: 'yaml',
+  HelmValues: 'yaml',
+  TerraformTfvars: 'tfvars',
+  TOML: 'toml',
+};
+
+type ConfigExtractorFormat = Extract<ConfigFormat, 'JSON' | 'Dotenv'>;
+
+type ExtractorParameters = { content: string };
+type ExtractorFunction = (params: ExtractorParameters) => Record<string, string>;
+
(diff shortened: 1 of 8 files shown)


packages/cli/src/helpers/configExtractor.ts
```diff
@@ -0,0 +1,100 @@
+import Dotenv from 'dotenv';
+import _ from 'lodash';
+import { Cfgu, Config } from '@configu/sdk';
+
+type ConfigFormat =
```
davidmeirlevy, line 5: that's the old implentation

packages/cli/src/helpers/configExtractor.ts
```diff
@@ -0,0 +1,100 @@
...
+  'JSON',
+  'Dotenv',
+]);
+
+const FILE_CONTENT_FORMAT_ERROR =
+  'file content is not in a valid format. for more info on supported formats read here: https://configu.com/docs';
+const NOT_FLAT_JSON_ERROR = 'only flat .json files are supported. read more here: https://configu.com/docs';
+
+const CONFIG_EXTRACTORS: Record<ConfigExtractorFormat, ExtractorFunction> = {
+  Dotenv: ({ content }) => Dotenv.parse(content),
```
davidmeirlevy, line 41: we should use the new integrations

davidmeirlevy (changes requested):

packages/cli/src/helpers/configExtractor.ts
RichardAkman: The integrations are not relevant, they are for stringification while this logic is for parsing, two different purposes.

packages/cli/src/helpers/configExtractor.ts
RichardAkman: The old implementation is what @pelegpor asked for. Once Ran is back, the `init` command will be reworked and this code will be changed accordingly.

packages/cli/src/helpers/configExtractor.ts
davidmeirlevy: Maybe we'll need another task to convert the usage here to the adaptive expressions, but that's another task.

davidmeirlevy (approved):

PR #644, opened 2024-10-31
Implement integration consumption

### LINKED ISSUE #642: Implement consumption of integrations by higher-level Configu interfaces
### Suggestion

Fully implement the logic that consumes the integrations packages in `@configu/integrations` in `@configu/cli` and `@configu/proxy`.

### Motivation

This is part of the core logic of `@configu/cli` and `@configu/proxy`.

### Context

_No response_

RichardAkman:
Closes #642

## Implementation notes
- Moved formatter integrations from `@configu/common` to `@configu/cli`
  - The registration of the formatters now happens as part of the CLI export command. This is because they are irrelevant to other commands. They are only registered if a formatting expression is evaluated.
  - The package dependencies of both`@configu/common` and `@configu/cli` were adjusted accordingly.
  - Some formatter integrations incorrectly exported the formatter functions, these packages were adjusted to make things work.
- Removed some `console.log` lines from `@configu/common` to clean up the codebase. These log lines cause the export command `Stdin` to be contaminated and result in `readPreviousEvalCommandOutput` failing the parse.
- Left the built-in integration as introduced by #550.
- Made adjustments to remote store registration:
  - `getStoreInstance` is now async to ensure that the store is registered before any logic that uses the registry to look up stores finds it. Adjusted all usages of this method in both `@configu/proxy` and `@configu/cli`.
  - Renamed `Registry.remoteRegister` to `remoteRegisterStore` since this logic is only relevant to store integrations. Expression integrations are built into the interface registry.
  - Added temporary commented-out logic that mapped the store type into a normalized string for the artifact look-up. Until the integration pipeline is updated by @rannn505, this won't work hence it being commented out. The same goes for the integration versioning logic.
- Implemented the caching logic by extracting the cache directory resolution from `@configu/cli` and moved it into `@configu/common` so it can be recycled
  - @rannn505, `clipanion` doesn't provide the cache directory logic that `oclif` did so I would appreciate your input on `getConfiguHomeDir` which attempt to replicate the cache directory resolution logic. I leverage some internal code I found within the `oclif` codebase. I followed your `$HOME/.configu/.cache` comment as the place to keep the cached integration files.


davidmeirlevy (approved):
Just a general note, prefer to add the "node:" prefix to internal node libraries, for security reasons.

packages/common/src/Registry.ts
```diff
@@ -4,8 +4,9 @@ import { mkdir, writeFile } from 'node:fs/promises';
 import path from 'pathe';
 import { tsImport } from 'tsx/esm/api';
 import { ConfigStore, ConfigStoreConstructor, Expression, ExpressionFunction } from '@configu/sdk';
+import { getConfiguHomeDir } from './utils';
 
-const CONFIGU_HOME = path.join(process.cwd(), '/.configu-cache');
+const CONFIGU_HOME = path.join(getConfiguHomeDir(), '/.cache');
```
RichardAkman, line 9: Ran's comment: rename cache to `integrations` or `stores`

packages/cli/src/commands/export.ts
```diff
@@ -16,8 +16,16 @@ import {
...
+import { readFile, Registry } from '@configu/common';
 import * as os from 'os';
+import { CompactJSON } from '@configu-integrations/compact-json';
+import { Dotenv } from '@configu-integrations/dotenv';
+import { HelmValues } from '@configu-integrations/helm-values';
+import { JSONExpression } from '@configu-integrations/json';
+import { jsonToKubernetesConfigMap } from '@configu-integrations/kubernetes-config-map';
+import { jsonToTfvars } from '@configu-integrations/terraform-tf-values';
+import { TOML } from '@configu-integrations/toml';
+import { YAML } from '@configu-integrations/yaml';
```
RichardAkman, line 28: Move into `registerFormatterExpressions`

packages/cli/src/commands/export.ts
```diff
@@ -155,6 +163,19 @@ export class CliExportCommand extends BaseCommand {
     return this.filterFromFlag(filteredConfigs, filterExpressions);
   }
 
+  async registerFormatterExpressions() {
+    await Registry.register({
```
RichardAkman, line 167: Ran's commnet: use `Registry.localRegister` with promise.all or `Registry.import`

packages/common/src/Registry.ts
RichardAkman: Decided to rename `cache` to `integrations` for future-proofing in case we ever decide to remote register something other than stores (expressions, etc..)

packages/common/src/utils.ts
```diff
@@ -65,3 +67,16 @@ export const parseYAML = (filePath: string, fileContent: string): any => {
     throw error;
   }
 };
+
+export const getConfiguHomeDir = (): string => {
```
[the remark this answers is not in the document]
RichardAkman: done, also changed the directory creation logic to use `fs/promises`.

RichardAkman:
Closed in favor of #659

PR #653, opened 2024-11-06
debug-locally: add debug configuration for proxy

michael-hoffman-26:
Adding local setup for debugging the Proxy service


rannn505 (approved):
@michael-hoffman-26 Looks good, thanks you :)

PR #658, opened 2024-11-12
Implement store docs generation script

### LINKED ISSUE #601: Create a script that mirrors store integration `REAME.md` files to docs in Configu docs website
### Suggestion

Following #600, all the `REAME.md` files should be leverage and converted into docs for store integrations under https://docs.configu.com/integrations.

### Motivation

Improves the completeness of the documentation websites and ensures a single source of truth for store documentation while leveraging existing content.

RichardAkman:
Closes #601


docs/integrations/index.json
RichardAkman, line 1: What do we do with the unused properties like the icon? @rannn505

rannn505 (approved):

PR #663, opened 2024-11-20
fix integrations bundle for different archs

davidmeirlevy:


.github/actions/ts-env-action/action.yml
```diff
@@ -27,7 +31,7 @@ runs:
       shell: bash
       if: ${{ inputs.install-dependencies == 'true' }} # * there is a bug in gh-actions causing booleans to be strings in composite actions - see: https://github.com/actions/runner/issues/1483#issuecomment-1042393939
       run: |
-        pnpm install
+        pnpm install --config.architecture=${{ inputs.architecture }}
```
rannn505, line 34: If the string is empty it is used the machine arch?

rannn505 (approved):

PR #668, opened 2024-11-25
Add backwards compatibility for legacy ConfigSchemas

RichardAkman:
Adds backward compatibility for the legacy ConfigSchema.

### DIFF
--- a/packages/sdk/src/ConfigSchema.ts
+++ b/packages/sdk/src/ConfigSchema.ts
@@ -3,6 +3,7 @@ import { CfguSchema } from './Cfgu';
 import { ConfigKey } from './ConfigKey';
 import { JSONSchema, JSONSchemaObject, FromSchema } from './expressions/JSONSchema';
 import { ConfigValue } from './ConfigValue';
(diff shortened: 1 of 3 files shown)


packages/sdk/src/ConfigSchema.ts
```diff
@@ -44,4 +90,63 @@ export class ConfigSchema {
...
+            }
+          });
+        }
+        if (labels) {
+          migratedCfgu.label = labels;
+        }
+        if (depends) {
+          depends.forEach((dependencyKey: string) => {
+            migratedCfgu.test.push(
+              `expect($.configs.${dependencyKey}.storedValue, 'Dependency ${dependencyKey} is missing for ${key}').to.exist`,
```
RichardAkman, line 119: Adjust to be only if current key exists

RichardAkman: Not relevant, EvalCommand does this on its own:

```
          if (origin !== EvaluatedConfigOrigin.Empty) {
            ConfigValue.validate({
              store,
              set,
              schema,
              current: key,
              configs: result,
            });
```

PR #695, opened 2025-01-06
Remove `config` naming restriction

### LINKED ISSUE #694: Remove `Config` naming restriction for backward compatibility
### Current Behavior

`Config` isn't an allowed keyword which causes backward compatibility issues

### Expected Behavior

`Config` Should be allowed

### To Reproduce

_No response_

### Context

_No response_

RichardAkman:
Closes #694


rannn505 (approved):
That's sad 😞

Issue #526, opened 2024-10-09
Enable SSE at `@configu/proxy`


rannn505:
### Suggestion

Enable Server-Sent Events (SSE) in `@configu/proxy`. The proxy should trigger SSE based on a cron job interval string passed as a query parameter to the `POST /export`, which runs the export again and sends back an SSE each time the interval hits.

### Motivation

This feature provides real-time updates via SSE for clients, improving responsiveness and enabling automated configuration exports at defined intervals or upon receiving hooks.

### Context

The SSE should respond either based on a cron job syntax provided by the user or when a hook is received at the proxy, ensuring configurations are exported and evaluated dynamically.

- https://en.wikipedia.org/wiki/Server-sent_events#:~:text=Server%2DSent%20Events%20(SSE),client%20connection%20has%20been%20established.
- https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events
- Use fastify <-> SSE module - https://fastify.dev/ecosystem/
<img width="1507" alt="image" src="https://github.com/user-attachments/assets/125f5664-66dd-4316-8418-a43109281a62">

rannn505:
@qballer did you had a chance to progress with this feature?

qballer:
Nope give it to someone else

BeGoCode:
1. @rannn505 Can you assign this to me?
2. Is this only a BE ticket? Or do we have a FE client we want to update accordingly as well?
3. Regarding tests, do we have any unit tests or any integration tests we usually do?

rannn505:
**@BeGoCode**  
1. Done! ✅  
2. The **@configu/proxy** is a standalone server that currently has one HTTP endpoint, but we’re planning to expand its capabilities to include SSE (this issue), WebSockets, GraphQL, and more. Its primary goal is to be deployed within the user’s environment and serve as the entry point for fetching configuration data at runtime. Therefore, a UI (or any frontend) is redundant in this context. That said, we’re exploring the idea of integrating a lightweight UI for all our interfaces (primarily for the CLI) to visualize the `.configu` local file and potentially enable configuration upserts through a UI.  
3. We had an extensive test suite, but following the major rewrite for v1, many tests were removed. We’re planning an internal "testing party" soon to restore and enhance the test coverage to 100%.
[...]

PR #697, opened 2025-01-14
Enable SSE at @configu/proxy

BeGoCode:
1. Introduced support for Server-Sent Events (SSE) in the proxy, enabling the use of a cron parameter to determine whether results are sent via a standard POST method or streamed continuously through SSE. Additionally, a cron job is activated to periodically fetch data when streaming is enabled.
2. Refactored the Fastify server initialization by moving it from the index file into a standalone module, allowing for the creation of independent server instances to facilitate testing.
3. Implemented test coverage for proxy endpoints.

Issue link: https://github.com/configu/configu/issues/526


packages/proxy/src/__tests__/proxy.test.ts
```diff
@@ -0,0 +1,167 @@
...
+          });
+        });
+      });
+
+      req.on('error', (err: string) => {
+        throw new Error(err);
+      });
+      req.write(body);
+      req.end();
+    });
```
davidmeirlevy, line 157: there's a better native way in node.js to create SSE request. here's an example:
```
const response = await fetch("/sse", { method: "POST" });
const reader = response.body.getReader();
const decoder = new TextDecoder("utf-8");
while (true) {
  const { value, done } = await reader.read();
  if (done) break;
  console.log("Received", decoder.decode(value));
}
```

packages/proxy/src/index.ts
```diff
@@ -95,13 +14,19 @@ export async function listen() {
...
-(async () => {
-  await listen();
-})();
+// Only run listen() if this file is the entry point
+const modulePath = fileURLToPath(import.meta.url);
+if (modulePath === process.argv[1]) {
+  (async () => {
+    await listen();
+  })();
+}
```
davidmeirlevy, line 32: this is not a good practice.
if you need the listen function outside for testing, then move the listen function outside of this file, and here you should execute it.

packages/proxy/src/routes.ts
```diff
@@ -50,8 +37,74 @@ const ok = {
...
+      const pipe = await previousResult;
+
+      const storeInstance = await ConfiguInterface.getStoreInstance(store);
+      if (!storeInstance) {
+        throw new Error(`store "${store}" not found`);
+      }
+
+      const setInstance = new ConfigSet(set);
+      // todo: fix this any
+      const schemaInstance = new ConfigSchema(keys as any);
```
davidmeirlevy, line 57: I keep asking myself if this code should be out of this function because the store instance and the set are the same for the entire event, so creating new instances for every value check may be inefficient.
My answer depends on the cronjob time frame.
for example: if the time is less than a minute, you might want to keep the reference, but if not - it's better to re-create them.

packages/proxy/src/routes.ts
```diff
@@ -65,58 +118,81 @@ export const routes: FastifyPluginAsync = async (server, opts): Promise<void> =>
...
-      return reply.send(exportRes);
+      // If there's no cron in the query, do the one-time export and return JSON
+      if (!request.query.cron) {
+        const exportRes = await runExportAndGetResult(request.body);
+
+        // TODO: consider if this is the right way to parse the result
+        // const parsedExportRes = JSON.parse(exportRes.result);
+
+        // Return one-time result
+        return reply.send(exportRes);
```
davidmeirlevy, line 129: in fastify, you can just return the value.

packages/proxy/src/routes.ts
```diff
@@ -65,58 +118,81 @@ export const routes: FastifyPluginAsync = async (server, opts): Promise<void> =>
...
+      // Otherwise, we are in SSE mode, triggered by the presence of a "cron" query parameter
+      if (!cron.validate(request.query.cron)) {
+        throw new Error(`Invalid cron expression: "${request.query.cron}"`);
+      }
+
+      reply.raw.writeHead(200, {
+        'Content-Type': 'text/event-stream',
+        'Cache-Control': 'no-cache',
+        Connection: 'keep-alive',
+      });
```
davidmeirlevy, line 141: use fastify api. e.g.
```
reply
    .code(200)
    .header('Content-Type', 'application/json; charset=utf-8')
```

davidmeirlevy (changes requested):
@BeGoCode Thanks for this PR and Feature.
It looks good.

I had some simple notes, and meanwhile, I will use our version to verify it's also working as expected.

Thanks!

packages/proxy/src/routes.ts
BeGoCode: AFAIK cron not go down to sub-minute resolutions so using it straightforward like this doesn't allow less than a minute

packages/proxy/src/routes.ts
BeGoCode: I'm pretty sure using its api doesn't allow sending the headers immediately (which is necessary in SSE).
Since it doesn't natively support SSE with its api since SSE is a lower-level HTTP feature we have to use `reply.raw` and this leads to using `reply.raw.writeHead`

packages/proxy/src/index.ts
BeGoCode: Good find this shouldn't be here.
In fact I'm not even exporting `listen` so I'll just remove this

packages/proxy/src/routes.ts
BeGoCode: Done

packages/proxy/src/__tests__/proxy.test.ts
```diff
@@ -0,0 +1,167 @@
+import assert from 'assert';
+import { test, before, after } from 'node:test';
+import path, { dirname } from 'path';
+import { fileURLToPath } from 'url';
+import { ConfiguInterface } from '@configu/common';
+import http, { IncomingMessage } from 'http';
```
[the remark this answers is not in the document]
BeGoCode: Done

packages/proxy/src/__tests__/proxy.test.ts
```diff
@@ -0,0 +1,167 @@
...
+      'Content-Type': 'application/json',
+    },
+  });
+
+  assert.strictEqual(response.statusCode, 200, 'Expected status code 200');
+  assert.strictEqual(typeof JSON.parse(response.body), 'object', 'Expected response body to be an object');
+});
+
+test('should handle SSE stream and close connection', async () => {
+  const query = querystring.stringify({ cron: '* * * * *' });
```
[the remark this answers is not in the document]
BeGoCode: Done

packages/proxy/src/__tests__/proxy.test.ts
BeGoCode: Very nice!
Wasn't familiar of this native way, I've implemented this

BeGoCode:
> @BeGoCode Thanks for this PR and Feature. It looks good.
> 
> I had some simple notes, and meanwhile, I will use our version to verify it's also working as expected.
> 
> Thanks!

Thanks for the comments, I fixed most and need to talk to you about 3 of them.
I sent you a discord DM

davidmeirlevy (approved):
@BeGoCode thanks for the PR.
It looks good!

BeGoCode:
@rannn505 We passed all reviews, can you merge please?

packages/proxy/src/__tests__/start.cfgu.json
RichardAkman, line 1: This cfgu file uses the previous deprecated format and should be updated.

BeGoCode: I updated the file using the new schema as used here: https://docs.configu.com/interfaces/.cfgu

packages/proxy/src/routes.ts
```diff
@@ -66,58 +125,81 @@ export const routes: FastifyPluginAsync = async (server, opts): Promise<void> =>
...
+          const exportRes = await runExportAndGetResult(request.body);
+
+          const data = JSON.stringify(exportRes);
+          const replyObj = {
+            id: getNextId(),
+            event: 'export',
+            data,
+          };
+
+          reply.raw.write(JSON.stringify(replyObj));
```
RichardAkman, line 171: I'm quite sure this is supposed to be something like this:
``` suggestion
          reply.raw.write(`data: ${JSON.stringify(replyObj)}\n\n`);
```

Otherwise, the events aren't being sent properly

packages/proxy/src/__tests__/start.cfgu.json
RichardAkman: Amazing

packages/proxy/src/routes.ts
BeGoCode: That's why I wrapped the response in an object, there's id, event ant data is preceded by `data:` followed by a newline.
This is an example of the response:
`{
   "id":1,
   "event":"export",
   "data":"{\"metadata\":{\"version\":\"0.0.0\",\"start\":4138.29…ING\":\"hey\",\"SUBJECT\":\"Change2\",\"MESSAGE\":\"\"}}"
}`

RichardAkman: Please check again, currently, the events aren't being sent, take a look at this video for reference:
https://github.com/user-attachments/assets/401c8071-78ff-405e-b14d-bcba81434402

BeGoCode: @RichardAkman Ah, I see. Replies in SSE should start with 'data:' and not have an object which holds a data property.
I've updated the code and the tests accordingly.

RichardAkman: Great, I tested again and I see the events coming through now. There's one other point I want to raise on this topic, I'm not seeing a good reason to have `exportRes` be sent stringified within the response object which is itself stringified. If there's a particular reason, I'd love to understand it. Otherwise, I think it should be sent as is so that the response body will look like this:
image
instead of this:
image

BeGoCode: @RichardAkman Yes the response definitely should look like the former.
Code updated accordingly.

RichardAkman: Amazing

RichardAkman (approved):
Looks great, thanks a lot for your contribution! We will be releasing this in the next version of `@configu/proxy`

PR #699, opened 2025-01-19
ConfiguModule

rannn505:
- flatten @configu/sdk to a single entry
- add npm v7 to common deps
- add install deps util
- add giget to common deps
- add template fetch util
- create a ConfiguModule class to handle module registration locally and remotely
- adjust ConfiguFile and other according to the change

Closes #


davidmeirlevy (approved):
LGTM

PR #711, opened 2025-02-16
Revert "Enable SSE at @configu/proxy"

rannn505:
Reverts configu/configu#697

It will be re-merged after next release


RichardAkman (approved):
LGTM

PR #712, opened 2025-02-16
v1 with stores

rannn505:
- common utils rewrite
- ConfiguModule - giget + @npmcli/arborist + jiti
- OutputStreams - block stdout, add debug log + color + table
- .configu interface config
- cli setup command initial version
- cli @clack/prompts


packages/cli/src/commands/setup.ts
```diff
@@ -0,0 +1,187 @@
...
+            output.push(`Open a new terminal to apply the changes.`);
+          }
+          return 'Bin added to PATH';
+        },
+      },
+    ]);
+
+    if (isGlobalOrVersionFlag) {
+      output.push(`Run \`configu --help\` to see the available commands.`);
+      output.push(`Run \`configu init\` to get started.`);
```
RichardAkman, line 175: The init command currently doesn't exist, maybe we should omit this until it returns to prevent confusing the user.

packages/schema/.configu.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "$id": "https://raw.githubusercontent.com/configu/configu/main/packages/schema/.configu.json",
+  "$id": "https://api.configu.com/schema/.configu.json",
```
RichardAkman, line 3: Are these schemas planned to be served via the platform API? Currently, the API doesn't serve these schemas.

RichardAkman (approved):
LGTM 🎖️

Take a look at the minor comments I left

davidmeirlevy (approved):
LGTM

packages/schema/.configu.json
rannn505: Thanks that was a bug!

PR #717, opened 2025-03-01
fix(formatters): export dotenv with special characters

orlevii:
If a string contains `#`, like `mysecret#password`, the dotenv file will be:

```dotenv
KEY=mysecret#password
```
This is incorrect because everything after `#` will be treated as a **comment** and **ignored** when the file is read.

Closes #


rannn505 (approved):
LGTM :)

PR #719, opened 2025-03-06
Export configmap convert numerical and boolean values to string first

### LINKED ISSUE #718: Export configmap should convert numerical and boolean values to string first
### Suggestion

Hi team, thanks for the nice project and we are trying to use configu to manage our configs in k8s deployments! However, we start seeing some issues when trying to adopt configu and this is a major one

Kubenetes config maps expect all values to be string, but configu has some type conversion to make numerical string into number and booleanic string to Boolean, but as the formatter is just putting the values as it is, so it will output config map with non string value. I could not find a way to stop this type convertion.

For example with this `example.cfgu.yaml`

```yaml
$schema: 'https://files.configu.com/schema/.cfgu.json'
keys:
  NUM:
    default: 123
  BOOL:
    default: true
  NUM_S:
    default: '123'
  BOOL_S:
    default: 'true'
  STRING:
    default: 'some-string'
  STRING_SPACE:
    default: 'some string'
```

and run `configu eval | configu export --format config-map`, I will expect it to give

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  creationTimestamp: 2025-03-06T15:35:24.271Z
data:
  NUM: '123'
  BOOL: 'true'
  NUM_S: '123'
  BOOL_S: 'true'
  STRING: some-string
  STRING_SPACE: some string
```

but actually, it is

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  creationTimestamp: 2025-03-06T15:35:24.271Z
data:
  NUM: 123
  BOOL: true
  NUM_S: 123
  BOOL_S: true
  STRING: some-string
  STRING_SPACE: some string
```

I have also try some work around - export to dotenv and use kubectl to create configmap from dotenv file, but this does not work. Reusing the same example, `configu eval | configu export --format dotenv > my.env` will give

```dotenv
NUM=123
BOOL=true
NUM_S=123
BOOL_S=true
STRING=some-string
STRING_SPACE="some string"
```

Then convert to config map using `kubectl create configmap my-config --from-env-file=my.env --dry-run -o yaml`

```yaml
apiVersion: v1
data:
  BOOL: "true"
  BOOL_S: "true"
  NUM: "123"
  NUM_S: "123"
  STRING: some-string
  STRING_SPACE: '"some string"'
kind: ConfigMap
metadata:
  creationTimestamp: null
  name: my-config
```

Looks like it can solve the above two points, have name and all values are string, but the value of STRING_SPACE now got extra quotes. It is due to k8s considers the quote in env file as part of the literal, and maintainer expects the env to have the quotes removed at the first place instead of assuming k8s will remove it. (Maybe can also consider removing the extra quotes in dotenv export?)

https://github.com/kubernetes-sigs/kustomize/issues/4525#issuecomment-1223288283

https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#create-configmaps-from-files

So the only "working" workaround for putting bool/number value to configmap is putting extra quotes

```yaml
$schema: 'https://files.configu.com/schema/.cfgu.json'
keys:
  NUM_SS:
    default: '"123"'
  BOOL_SS:
    default: '"true"'
```

can give 

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  creationTimestamp: 2025-03-06T15:50:18.155Z
data:
  NUM_SS: "123"
  BOOL_SS: "true"
```

However this workaround will leads to extra quote when outputting in other format, which I dont think is a proper way to solve the problem. So to really make the configmap export work nicely, I think configu need to convert the values to string before putting them into the yaml file, as no one would expect number and boolean value in a config map.

I have a quick implementation for this, will submit PR shortly.

### Motivation

_No response_

### Context

_No response_

dipsywong98:
Closes #718

I have a test for the `ensureStringValue` passing but not sure where to put it. And I am facing issue running development environment so didn't really try out the function end to end, need some help to verify

```ts
describe('ensureStringValue', () => {
  it('t', () => {
    expect(ensureStringValue({
      nil: null,
      undefined: undefined,
      num: 123,
      bool: true,
      string: 'hello world'
    })).toEqual({
      nil: null,
      undefined: undefined,
      num: '123',
      bool: 'true',
      string: 'hello world'
    })
  })
})
```

### DIFF
--- a/packages/formatters/src/YAML.ts
+++ b/packages/formatters/src/YAML.ts
@@ -7,6 +7,20 @@ type YAMLFormatterOptions = Exclude<Parameters<typeof stringify>['2'], string | 
 export const YAMLFormatter: FormatterFunction = (configs, options?: YAMLFormatterOptions) =>
   stringify(configs, options);
 
+const ensureStringValue = (config: { [key: string]: unknown }) =>
+  Object.fromEntries(
+    Object.entries(config).map(([key, value]) => {
+      switch (typeof value) {
+        case 'number':
+          return [key, value.toString()];
+        case 'boolean':
+          return [key, value.toString()];
+        default:
+          return [key, value];
+      }
+    }),
+  );
+
 const DEFAULT_API_VERSION = 'v1';
 const DEFAULT_KIND = 'ConfigMap';
 export const KubernetesConfigMapFormatter: FormatterFunction = (
@@ -21,7 +35,7 @@ export const KubernetesConfigMapFormatter: FormatterFunction = (
       ...options?.metadata,
       name: options?.name?.toLowerCase(),
     },
-    data: configs,
+    data: ensureStringValue(configs),
   };
 
   return YAMLFormatter(jsonConfigMap);


rannn505:
hi @dipsywong98, thanks for this amazing contribution, could you elaborate more about those troubles running development environment? did you try dev container?

dipsywong98:
Hi @rannn505 thanks for the reply. Yes, I am using dev container and I ran `pnpm install`, then I am using `pnpm configu` to run the development version of configu as said in https://github.com/configu/configu/blob/main/packages/cli/README.md. However it looks outdated that in latest version configu we do `configu eval | configu export --format config-map` and in devcontainer `pnpm configu eval | pnpm configu export --format config-map` gives

```
 ›   Error: The following error occurred:
 ›     Missing required flag schema
 ›   See more help with --help
 ›   Error: Expected --format=config-map to be one of: JSON, CompactJSON, YAML, Dotenv, KubernetesConfigMap, HelmValues, TerraformTfvars, TOML
 ›   See more help with --help
```

and `pnpm configu --version` gives `@configu/cli/0.17.1 linux-x64 node-v20.10.0`. 

The devcontainer is built from commit 1cd9d85ac252c68232aff7d6b4757bdbf3c0b0d5

`.devcontainer/bootstrap.sh` does not exist as well although it is stated in https://github.com/configu/configu/blob/main/CONTRIBUTING.md#local-environment

PR #721, opened 2025-03-09
feat(cli): better docker image

orlevii:
Closes #

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+.dockerignore
+.git
+.github
+**/Dockerfile
+examples
+.vscode
--- a/packages/cli/Dockerfile
+++ b/packages/cli/Dockerfile
@@ -8,15 +8,20 @@ LABEL author "dev@configu.com"
 # https://pnpm.io/docker#example-2-build-multiple-docker-images-in-a-monorepo
 ENV PNPM_HOME="/pnpm"
 ENV PATH="$PNPM_HOME:$PATH"
+ARG COREPACK_VERSION="0.32.0"
 # https://vercel.com/guides/corepack-errors-github-actions
-RUN npm install -g corepack@latest
+RUN npm install -g "corepack@${COREPACK_VERSION}"
 RUN corepack enable
 RUN pnpm --version
 
 FROM base AS build
-COPY --exclude=**/node_modules --exclude=**/*.md --exclude=**/build --exclude=**/dist --exclude=**/tmp . /src
-WORKDIR /src
 
+WORKDIR /src
+COPY pnpm-lock.yaml ./
+COPY --parents packages/**/package.json patches ./
+RUN pnpm fetch
+
+COPY --exclude=**/node_modules --exclude=**/*.md --exclude=**/build --exclude=**/dist --exclude=**/tmp . /src
 RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install --frozen-lockfile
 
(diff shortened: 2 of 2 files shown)


.dockerignore
```diff
@@ -0,0 +1,6 @@
+.dockerignore
+.git
+.github
+**/Dockerfile
+examples
+.vscode
```
orlevii, line 6: This makes it more likely we hit the cache

packages/cli/Dockerfile
```diff
@@ -8,32 +8,45 @@ LABEL author "dev@configu.com"
...
 USER node
 
 ENV CONFIGU_HOME="$HOME/.configu"
 ENV CONFIGU_BIN="$CONFIGU_HOME/bin"
-COPY --from=build --chown=node:node /dist $CONFIGU_BIN
-WORKDIR $CONFIGU_BIN
+ENV PATH="$CONFIGU_BIN:$PATH"
+COPY --from=build --chown=node:node /dist/dist/configu.cjs $CONFIGU_BIN/configu
 
+WORKDIR /app
```
rannn505, line 44: how about going with pwd instead of /app?
https://docs.docker.com/reference/cli/docker/container/run/#volume

```bash
docker  run  -v $(pwd):$(pwd) -w $(pwd) -i -t  ubuntu pwd
```

rannn505 (approved):
@orlevii this is an awesome work, we shall replicate it afterwords to the proxy and also apply further optimizations like splitting arch between gh-action workers etc.

packages/cli/Dockerfile
orlevii: Confirmed to be working without, updating

packages/cli/Dockerfile
```diff
@@ -8,32 +8,44 @@ LABEL author "dev@configu.com"
...
 
 FROM base AS build
-COPY --exclude=**/node_modules --exclude=**/*.md --exclude=**/build --exclude=**/dist --exclude=**/tmp . /src
+
 WORKDIR /src
+COPY pnpm-lock.yaml ./
+COPY --parents packages/**/package.json patches ./
+RUN pnpm fetch
 
+COPY --exclude=**/node_modules --exclude=**/*.md --exclude=**/build --exclude=**/dist --exclude=**/tmp . /src
```
davidmeirlevy, line 24: wouldn't it be bette rto exclude node modules using the dockerignore?

packages/cli/Dockerfile
```diff
@@ -8,32 +8,44 @@ LABEL author "dev@configu.com"
...
 ENV CONFIGU_HOME="$HOME/.configu"
 ENV CONFIGU_BIN="$CONFIGU_HOME/bin"
-COPY --from=build --chown=node:node /dist $CONFIGU_BIN
-WORKDIR $CONFIGU_BIN
+ENV PATH="$CONFIGU_BIN:$PATH"
+COPY --from=build --chown=node:node /dist/dist/configu.cjs $CONFIGU_BIN/configu
 
 ENV NODE_ENV="production"
-ENTRYPOINT [ "node", "dist/configu.cjs" ]
+ENTRYPOINT [ "configu" ]
```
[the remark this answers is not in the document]
rannn505: it uses the js file with a `node shebang`

packages/cli/Dockerfile
rannn505: updated the example comment in the dockerfile

PR #747, opened 2025-08-19
Upgrade Node.js to LTS

rannn505:
- pnpm-specific configuration has been removed from .npmrc and added to pnpm-workspace.yaml.
- The prepare script now updates the Node.js version in pnpm-workspace.yaml instead of .npmrc
- Node.js upgraded to v22.18.0


RichardAkman (approved):
LGTM

